PR #52, opened 2019-01-31
Fix issues with company no. validation

Cruikshanks:
It was spotted during the testing of the new service that if you entered a registration number using lowercase characters, or added any whitespace then the check against companies house would fail.

The registration form should handle this on behalf of users rather than returning an error and expecting them to correct it. The changes included here will resolve both problems.

### DIFF
--- a/app/forms/waste_exemptions_engine/registration_number_form.rb
+++ b/app/forms/waste_exemptions_engine/registration_number_form.rb
@@ -28,10 +28,10 @@ module WasteExemptionsEngine
     private
 
     def process_company_no(company_no)
-      number = company_no.to_s
+      number = company_no.to_s.strip
       # Should be 8 characters, so if it's not, add 0s to the start
       number = "0#{number}" while number.length < 8
-      number
+      number.upcase
     end
   end
 end
--- a/spec/factories/forms/registration_number_form.rb
+++ b/spec/factories/forms/registration_number_form.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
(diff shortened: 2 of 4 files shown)


app/forms/waste_exemptions_engine/registration_number_form.rb
```diff
@@ -28,10 +28,10 @@ def submit(params)
     private
 
     def process_company_no(company_no)
-      number = company_no.to_s
+      number = company_no.to_s.strip
```
irisfaraway: Is this duplicating work which the CanStripWhitespace concern should be doing? I think all attributes should be stripped before validation.

Cruikshanks: Yes, you are right. Wonder why it's not picking it up. Will investigate and come back with edits.

Cruikshanks: Ahhhh! Now know why its not picking it up. `CanStripWhitespace` is a concern added to the `BaseForm`. Each form will sort out its own attributes (in this case it's just `company_no`) and then pass them to the base `submit()` method. In there is passes them to the concern's `strip_whitespace()` method.

All good so far, we're ensuring the values we save to the database are stripped of whitespace. The issue I've realised is we never update the properties of the form afterwards, which were originally used to create the attributes. And it's the form and its properties all validators look at.

This is why the value we save is stripped, but the value we validate is not.

As our intent is to strip whitespace before both saving and validating, I think this will require a bit more thought before we do anything as to what the solution will be, because all existing form objects are affected by this in both services (WCR & WEX).

PR #53, opened 2019-01-31
Make choose exemptions checkboxes more accessible

Cruikshanks:
It was noted on the existing service that the accessibility of the checkboxes could be improved, specifically how they appear to screen readers.

This also applied to this version. Prior to this change the screen reader would just read the code for the checkbox. So a user tabbing through would just hear "U1. U2. U3. Selected U3".

With this change they will now also hear the summary other users can see visually on the page e.g. "U1 Use of waste in construction. U2 Use of baled end-of-life tyres in construction".


app/views/waste_exemptions_engine/exemptions_forms/new.html.erb
```diff
@@ -30,7 +30,7 @@
               <td>
                 <div class="multiple-choice">
                   <%= f.check_box :exemptions, { :multiple => true, id: "exemptions_form_checkbox-#{exemption.code}", checked: @exemptions_form.matched_exemption_ids.include?(exemption.id) }, exemption.id, nil %>
-                  <%= f.label "checkbox-#{exemption.code}", exemption.code %>
+                  <label for="exemptions_form_checkbox-<%= exemption.code %>"><%= exemption.code %><span class="visually-hidden"> <%= exemption.summary %></span></label>
```
irisfaraway: It's possible to do this using the `f.label` block syntax: https://github.com/DEFRA/waste-exemptions-engine/blob/8266da8ede9c939822f0b49b4466c4b82d1b56ee/app/views/waste_exemptions_engine/shared/_manual_address.html.erb#L11-L14

Think we should do that over raw HTML for consistency's sake.

Cruikshanks: Ooh, awesome. Just my lack of knowledge there. Will update.

PR #60, opened 2019-02-05
Fix transpose farm answers in check answers page

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-99

An issue with the current version of the service is that the answers to whether the waste activity will

- take place on a farm
- the person responsible is a farmer

get transposed in the check your answers screen. When we migrated the functionality to this engine, we also inadvertently copied this same transposing issue.

The root cause is the bad naming of the routes. Prior to this change they were

- `is_a_farm` meant is the activity taking place on a farm
- `on_a_farm` meant is the person responsible a farmer

It's easy to see why with names like this the information gets transposed. So this change resolves the root problem by renaming the routes, and all related objects (controllers, forms, views, column names etc), and ensures the correct info is displayed with the correct title in the check your answers page.


Cruikshanks:
I know CodeClimate is giving me the slap down, but I thought it was cleaner to leave the specs out whilst @eminnett is doing his 'thang'. At the very least total coverage remains unaffected 😬 .

app/views/waste_exemptions_engine/on_a_farm_forms/new.html.erb
```diff
@@ -11,7 +11,7 @@
     <% else %>
     <div class="form-group">
     <% end %>
-      <fieldset class="inline" id="on_a_farm">
+      <fieldset class="inline" id="is_a_farm">
```
irisfaraway: Should this have been changed?

config/locales/forms/check_your_answers_forms/en.yml
```diff
@@ -43,12 +43,12 @@ en:
...
         farming:
           subheading: Farming information
-          is_a_farm:
-            true_html: "the operater <span class=\"strong\">is</span> a farmer"
-            false_html: "the operater <span class=\"strong\">is not</span> a farmer"
           on_a_farm:
             true_html: "the operation <span class=\"strong\">will</span> take place on a farm"
             false_html: "the operation <span class=\"strong\">will not</span> take place on a farm"
+          is_a_farmer:
+            true_html: "the operater <span class=\"strong\">is</span> a farmer"
```
irisfaraway: I know this was an old issue, but the spelling is actually `operator`

app/views/waste_exemptions_engine/on_a_farm_forms/new.html.erb
eminnett: Well caught.

Cruikshanks: 🤦‍♂️

config/locales/forms/check_your_answers_forms/en.yml
Cruikshanks: 🤦‍♂️

eminnett:
I can't see anything other than what Iris has already commented on. It's possible I might find something when I write the tests, but I'll fix it then if I do.

PR #61, opened 2019-02-06
Fix postcode form validation

eminnett:
https://eaflood.atlassian.net/browse/RUBY-83

The postcode forms fail to correctly handle postcodes that match zero addresses. The expected behaviour is that the form validation displays an error asking the user to either correct the postcode or manually enter an address. This matches the behaviour on Waste Carriers.

This work includes additional changes to the automated tests both to setup what is required to test the postcode form but also to support upcoming work on the automated tests.


eminnett:
I refactored the PostcodeValidator tests as we discussed (only test the public interface).

spec/validators/waste_exemptions_engine/postcode_validator_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+            .to receive(:validate_each)
+            .once
+
+          validatable.valid?
+        end
+
+        it "the errors are empty" do
+          validatable.valid?
+          expect(validatable.errors).to be_empty
+        end
```
Cruikshanks: I think the one additional test I'd like to see is that the object we have added the validator too is valid.

```ruby
it "confirms the object is valid" do
  expect(validatable).to be_valid
end
```

At the moment it feels like we are assuming this from other test results rather than directly checking.

spec/validators/waste_exemptions_engine/postcode_validator_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      end
+
+      context "when the postcode is not valid" do
+        subject(:validatable) { Test::PostcodeValidatable.new(invalid_postcode) }
+
+        it "adds a validation error to the record" do
```
Cruikshanks: I think the reason why this is invalid would be made clearer (especially in the output) if this was moved into the context defined at line 57.

spec/validators/waste_exemptions_engine/postcode_validator_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+          validatable.valid?
+          expect(validatable.errors[:postcode].count).to eq(1)
+        end
+
+        context "because the postcode is not present" do
+          subject(:validatable) { Test::PostcodeValidatable.new }
+
+          it "adds an appropriate validation error" do
+            validatable.valid?
+            expect(validatable.errors[:postcode]).to eq(["Enter a postcode"])
```
Cruikshanks: I'd advocate referring back to the locale rather than hard coding the expected text in the test.

Experience on digital services is that content changes frequently so hard coding tends to make things quite brittle. However this one probably needs more of a chat. This might be my lack of experience showing through 😁

spec/validators/waste_exemptions_engine/postcode_validator_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+          validatable.valid?
+        end
+
+        it "the errors are empty" do
+          validatable.valid?
+          expect(validatable.errors).to be_empty
+        end
+      end
+
+      context "when the postcode is not valid" do
```
Cruikshanks: Again I'd be more comfortable if we included tests that check the object the validator has been added to is showing as not valid in these contexts.

spec/dummy/config/initializers/waste_exmeptions_engine.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
eminnett: Fixed in the latest commit.

spec/validators/waste_exemptions_engine/postcode_validator_spec.rb
eminnett: Fixed in the latest commit.

spec/validators/waste_exemptions_engine/postcode_validator_spec.rb
eminnett: Fixed in the latest commit.

spec/validators/waste_exemptions_engine/postcode_validator_spec.rb
eminnett: Fixed in the latest commit.

spec/validators/waste_exemptions_engine/postcode_validator_spec.rb
eminnett: Fixed in the latest commit.

PR #62, opened 2019-02-08
Inherit authorisation/authentication preferences from host

irisfaraway:
Improved testing during DEFRA/waste-exemptions-back-office-ta#21 revealed that the mounted engine wasn't using any of the auth checks provided by Devise in the host app - an unauthenticated user couldn't access routes from the host app, but could access ones in the mounted engine.

This also applied to checks for abilities as well.

The front-office doesn't use Devise or CanCanCan to manage authorisation or authentication, but the engine should be able to handle it when the host app specifies it.

### DIFF
--- a/app/controllers/waste_exemptions_engine/application_controller.rb
+++ b/app/controllers/waste_exemptions_engine/application_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module WasteExemptionsEngine
-  class ApplicationController < ActionController::Base
+  class ApplicationController < ::ApplicationController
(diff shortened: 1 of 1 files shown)


irisfaraway:
@Cruikshanks Spent ages trying to get https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/21 to work without modifying the engine at all but it was not happening. So made this change after reading https://stackoverflow.com/questions/12208448/how-to-use-devise-cancan-to-protect-mounted-engine-resources.

We didn't run into this issue with waste carriers as both host apps were making these sorts of checks, so we could just build it straight into the engine.

PR #65, opened 2019-02-11
Refactor AddressValidator

eminnett:
In an effort to remove duplication in the validators, we would like to move duplicated code into the ValidatesPresence concern and BaseValidator class. 

Future refactoring efforts will include the creation of the ValidatesLength concern.

This PR is a sample of the intended refactoring. We decided it would be easier to review the changes if they are applied to a single validator before applying the pattern to the rest of the validators.

### DIFF
--- a/app/validators/concerns/waste_exemptions_engine/validates_presence.rb
+++ b/app/validators/concerns/waste_exemptions_engine/validates_presence.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module WasteExemptionsEngine
+  module ValidatesPresence
+    extend ActiveSupport::Concern
+
+    included do
+      private
+
+      def value_is_present?(record, attribute, value)
+        return true if value.present?
+
+        record.errors[attribute] << error_message(record, attribute, "blank")
+        false
+      end
+    end
+  end
+end
--- a/spec/support/shared_examples/validators/presence_validator.rb
+++ b/spec/support/shared_examples/validators/presence_validator.rb
@@ -7,8 +7,8 @@ RSpec.shared_examples "a presence validator" do |validatable_class, property|
   # end
 
   describe "#validate_each" do
-    context "when the postcode is not valid" do
-      context "because the postcode is not present" do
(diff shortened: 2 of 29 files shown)


app/validators/concerns/waste_exemptions_engine/validates_presence.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module WasteExemptionsEngine
+  module ValidatesPresence
```
Cruikshanks: It's a name thing! If you look in `app/models/concerns/waste_exemptions_engine` you'll see we have a convention of naming our concerns `CanDoThing`. So in this case something along the lines of `CanValidatePresence` would bring it into line with that convention.

spec/support/shared_examples/validators/presence_validator.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+RSpec.shared_examples "a presence validator" do |validatable_class, property|
+  # it "includes ValidatesPresence" do
```
Cruikshanks: Another convention we have is not to leave commented out code in the code base.

> I should really write these down somewhere 😳

eminnett: This was an oversight.  It shouldn't have been commented out.

app/validators/concerns/waste_exemptions_engine/validates_presence.rb
eminnett: I'll fix this now.

spec/support/shared_examples/validators/presence_validator.rb
eminnett: Fixed in bcdf764.

app/validators/concerns/waste_exemptions_engine/validates_presence.rb
eminnett: Fixed in bcdf764

PR #69, opened 2019-02-15
Fix address form 500 error

eminnett:
https://eaflood.atlassian.net/browse/RUBY-83

While testing RUBY-83, Tim found an unhandled 500 error when submitting the address form. This PR attempts to fix that error.

WIPThe includes includes an adjustment of the validation in the AddressLookupForm class to ensure the messaging and behaviour is consistent with WCR.


eminnett:
I am submitting this for review despite Codeclimate failing the last commit on the grounds of code coverage. The tests that will cover this PR are being written as a part of PR #70.

PR #70, opened 2019-02-15
Add address form tests

eminnett:
Adding the automated tests to WEX is being broken down into chunks. This is the address form tests chunk.


eminnett:
@irisfaraway I've decided to refactor the AddressManualForm validators into a custom validator in a separate PR. I also realised that the existing form specs are missing tests that ensure that the forms are updating the transient registration as expected. I have added this for the address form tests in this PR, but will update the other form specs in the next form tests PR (where the remaining specs will be added).

PR #75, opened 2019-02-20
Add remaining form tests

eminnett:
The form tests were broken down into multiple chunks. The address form tests have already been written. This PR applies the same pattern to exercise the remaining forms.


spec/forms/waste_exemptions_engine/operator_name_form.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
```
irisfaraway: This filename is missing `_spec` at the end, so RSpec isn't running it and it gets missed in the code coverage.

irisfaraway: This is actually the case for `operator_postcode_form.rb` and `site_postcode_form.rb` - not from this PR but I just noticed they're missing the `_spec` as well. I think that should bump the coverage up!

spec/forms/waste_exemptions_engine/location_form_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+module WasteExemptionsEngine
+  RSpec.describe LocationForm, type: :model do
+    subject(:form) { build(:location_form) }
+
+    LOCATIONS = %w[
+      england
+      northern_ireland
+      scotland
+      wales
+    ].freeze
```
irisfaraway: Just occurred to me – if we have `frozen_string_literal: true` at the top, do we need to be explicitly freezing `LOCATIONS` and others like it? I don't think it hurts, but also my understanding is that the comment makes it unnecessary.

eminnett: I think Rubocop is doing this. I'll quickly check.

spec/forms/waste_exemptions_engine/operator_name_form.rb
eminnett: Thanks for noticing. I'll fix this in a moment.

eminnett: This is fixed in c0b9b1c.

irisfaraway: I noticed because I've done exactly the same thing!

PR #79, opened 2019-02-22
Make find_address_by_type work with reg addresses

irisfaraway:
This came out of https://eaflood.atlassian.net/browse/RUBY-60.

This commit should allow us to call `find_address_by_type` on a registration and return the correct addresses. Previously it would only check related transient_addresses, which a registration should never have.


irisfaraway:
Would have written specs for this but wasn't sure if that'd clash with https://eaflood.atlassian.net/browse/RUBY-196 – thoughts?

irisfaraway:
Didn't end up needing this so gonna go ahead and close it.

PR #81, opened 2019-02-25
Add models tests

eminnett:
The automated tests are being added in chunks. This is the models chunk.


app/models/concerns/waste_exemptions_engine/can_have_registration_attributes.rb
```diff
@@ -14,23 +14,23 @@ def partnership?
...
     def site_address
-      find_address_by_type(TransientAddress.address_types[:site])
+      find_address_by_type("site")
     end
 
     private
 
     def find_address_by_type(address_type)
-      return nil unless transient_addresses.present?
+      return nil unless addresses.present?
```
irisfaraway: Excited to see this in here as I was about to do the same thing for the reg/t_reg overview page! Woop 👏

eminnett: Yes, when I saw your PR for the aliased associations I new it would help tidy up that concern.

PR #84, opened 2019-02-26
Override methods for submitting dead-end forms

irisfaraway:
We have some 'forms' which aren't really forms as the user can't submit them. They are only used when the user journey reaches a dead end (at least within the service).

This commit overrides the `submit` methods in the form classes and the `create` methods in the form controllers so that we don't attempt to go through the usual process of updating a transient_registration with form data. This is consistent with how we handle these forms in waste carriers, and how the RegistrationCompleteForm works within this engine.


irisfaraway:
No test coverage as tests for these forms are planned to be in a separate PR, and we have no request specs yet.

PR #86, opened 2019-02-27
Implement PaperTrail for auditing

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-203

PaperTrail is a gem used for keeping a record of when and how a record is changed. We use it to keep track of changes to registrations. This information will then be shown in the WEX back office.

This PR will install and configure PaperTrail so we can migrate our existing audit history and continue to build on it.


irisfaraway:
Not ready for review, just need to update the branch!

spec/requests/waste_exemptions_engine/registration_complete_forms_spec.rb
```diff
@@ -18,6 +18,29 @@ module WasteExemptionsEngine
...
+        end
+
+        it "assigns the correct whodunnit to the registration version", versioning: true do
+          get request_path
+          registration = WasteExemptionsEngine::Registration.where(reference: form.transient_registration.reference).first
+          expect(registration.reload.versions.last.whodunnit).to eq(current_user.id.to_s)
+        end
+      end
+
+      context "when the does not have a current_user" do
```
Cruikshanks: Guessing we're missing something here 😁 

```suggestion
      context "when the host application does not have a current_user" do
```

PR #92, opened 2019-03-11
Update VCR cassettes

irisfaraway:


irisfaraway:
Trying to get the script to work with this repo as well as waste-carriers-renewals, but clearly something has gone wrong with this one as the cassettes were not refreshed.

PR #93, opened 2019-03-11
Refactor 'can navigate flexibly' form property

eminnett:
We decided to refactor the `CanNavigateFelxibly` concern into a class method on the BaseForm. The class method can then be overridden in subclasses if that specific form should not be allowed to navigate flexibly.


eminnett:
I don't appear to be able to do anything about the 12 'new' issues in code climate. Either I don't have permission to mark them as won't fix or we have to wait until the PR is merged to mark them as won't fix.

PR #98, opened 2019-03-14
Save the assistance mode for a registration

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-200

A registration should have a different assistance_mode depending on whether it was submitted through the front-office or back-office. This allows us to track what proportion of registrations are assisted digital.


irisfaraway:
Oh hey, service tests! These definitely needs some more work to properly cover everything that's going on, but I think that should be done in a separate PR for clarity's sake.

PR #100, opened 2019-03-15
Load engine migrations before host app migrations

irisfaraway:
This came out of our story to implement PaperTrail in this engine and the back office. Since we were tracking models in both projects, we ended up with two migrations to create a Versions table, which then caused a clash. So rather than have two migrations, it makes sense for the engine's migrations to take precidence, and for it to handle the creation of the versions table.


irisfaraway:
Hoping this will fix the issue from https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/89

PR #105, opened 2019-03-21
Refactor CanChangeExemptionStatus into CanActivateExemption

eminnett:
We decided to limit the scope of `CanChangeExemptionStatus` in the WEX engine to only handle activations. Deregistrations will be handled through a service in the back-office repository.

### DIFF
--- a/app/models/concerns/waste_exemptions_engine/can_activate_exemption.rb
+++ b/app/models/concerns/waste_exemptions_engine/can_activate_exemption.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module WasteExemptionsEngine
-  module CanChangeExemptionStatus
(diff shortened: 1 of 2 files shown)


spec/support/shared_examples/models/changer_of_the_exemption_status.rb
```diff
@@ -1,13 +1,12 @@
 # frozen_string_literal: true
```
irisfaraway: Should this file be renamed to `activator_of_exemptions.rb`?

eminnett: Yes it should.

eminnett: This is fixed in 024adc8.

PR #112, opened 2019-03-29
Add email interception and retrieval

eminnett:
In order to improve the flow of the end-to-end tests, we want to add behaviour to the application that will support the storage of the last email that was sent. There will also be a route that returns the latest mail object as JSON. The feature will be enabled through the use of an ENV var so we can ensure the endpoint is only accessible when the feature is explicitly enabled.

### DIFF
--- a/app/controllers/waste_exemptions_engine/last_email_controller.rb
+++ b/app/controllers/waste_exemptions_engine/last_email_controller.rb
@@ -3,17 +3,7 @@
 module WasteExemptionsEngine
   class LastEmailController < ApplicationController
     def show
-      if ENV["ENABLE_LAST_EMAIL_CACHE"] == "true"
-        render json: WasteExemptionsEngine::LastEmailCacheService.instance.last_email_json
-      else
-        handle_disabled
-      end
-    end
-
-    private
-
-    def handle_disabled
-      redirect_to error_path(404), status: 404
+      render json: WasteExemptionsEngine::LastEmailCacheService.instance.last_email_json
     end
   end
 end
--- a/app/controllers/waste_exemptions_engine/last_email_controller.rb
+++ b/app/controllers/waste_exemptions_engine/last_email_controller.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module WasteExemptionsEngine
+  class LastEmailController < ApplicationController
+    def show
+      if ENV["ENABLE_LAST_EMAIL_CACHE"] == "true"
+        render json: WasteExemptionsEngine::LastEmailCacheService.instance.last_email_json
+      else
+        handle_disabled
+      end
+    end
+
+    private
+
+    def handle_disabled
+      redirect_to error_path(404), status: 404
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


app/controllers/waste_exemptions_engine/last_email_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module WasteExemptionsEngine
+  class LastEmailController < ApplicationController
+    def show
+      if ENV["ENABLE_LAST_EMAIL_CACHE"] == "true"
```
irisfaraway: We have a page in waste carriers which we only use for testing purposes, but we have the logic to enable/disable it in the route instead: https://github.com/DEFRA/waste-carriers-renewals/blob/master/config/routes.rb#L5 That feels a bit simpler to me but I suppose you could argue that it clutters up the routes file. What do you think?

Cruikshanks: Think I'd rather keep to the convention that config values are decided and configured by the host application rather than the engine referring to env vars directly. 

https://github.com/DEFRA/waste-exemptions-engine/blob/master/lib/waste_exemptions_engine.rb

Appreciate a bit like `USE_XVFB_FOR_WICKEDPDF` it doesn't really matter where the engine reads this value from, but for consistency I think it best to stick to the convention of host app reads the env var and then configures the engine via its config block in an initialiser.

app/controllers/waste_exemptions_engine/last_email_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+        render json: WasteExemptionsEngine::LastEmailCacheService.instance.last_email_json
+      else
+        handle_disabled
```
Cruikshanks: `handle_disabled` is a one line method. As such I think we can just include that one line here, but there may be a case or reason I am not seeing.

app/controllers/waste_exemptions_engine/last_email_controller.rb
Cruikshanks: Think I prefer the routes version, and that answers my question regards `handle_disabled` 😁

eminnett: Fixed in 15b3b13.

app/controllers/waste_exemptions_engine/last_email_controller.rb
eminnett: Fixed in 15b3b13.

app/controllers/waste_exemptions_engine/last_email_controller.rb
eminnett: Fixed in 15b3b13.

eminnett:
While making these changes, I discovered a bug with the engine routing. The way the error route is written means all requests that should 404 get swallowed up by the errors route and result in a 500.

I have discussed this with Alan and we will discuss the problem in the next dev catchup. In the meantime, I have marked the tests that confirm the 404 as pending.

PR #117, opened 2019-04-03
Add rake task to generate privacy policy PDF

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-216

There are no doubt online tools that could take html and generate a PDF for us, however we have already built into the engine the ability to generate PDF's from html. So why not just use that funcationality? 🤔💡

Hence this change which adds a html version of the privacy policy, and a rake task to convert that to a pdf file. It outputs to a tmp folder to avoid the file getting accidently commited here. The intention is that you generate the file when needed here, and then simply copy and past it to the apps that host the engine.

As we expect to do this so infrequently, we decided this simple approach was preferable to trying to do anything to0 clever in the code.

---

WIP


Cruikshanks:
Think I am going to scratch this altogether. I intended to upload both versions of the PDF into the relevant Jira story and leave the final decision as to which one to pursue with our QA and Product Manager.

However when clicking on the attachments in Jira via Chrome I spotted that

- the links in the online generated one worked
- the links in the wicked_pdf generated one had stopped working again

There is clearly something that is not being done properly when adding the links to the html that I have come round to the decision that we should close this PR, salvage what we can from it (see PR #118 and #119 ) and just use version generated by https://html2pdf.com/

Cruikshanks:
Just could not get the links in the generated PDF to work, and it's so easy to just upload the privacy_policy.html to an online generator and get a working PDF that it doesn't seem sensible to add this additional code to the project.

PR #119, opened 2019-04-03
Add html view which is used to generate PDF privacy

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-216

The need to generate and update the PDF will be very infrequent (we are talking years) so rather than try and do anything ourselves we will just take advantage of available online generators.

For them to work they need to be provided with the HTML to convert, hence this change which adds pdf specific version of the privacu policy we can upload to be converted.

N.B. At this time (April 2019) we are using https://html2pdf.com/ to convert the html to PDF.

### DIFF
--- a/app/views/waste_exemptions_engine/pdfs/privacy_policy.html
+++ b/app/views/waste_exemptions_engine/pdfs/privacy_policy.html
@@ -1,0 +1,139 @@
+<!DOCTYPE html>
+<html>
+  <head>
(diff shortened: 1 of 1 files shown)


Cruikshanks:
For reference here is the pdf version

[privacy_policy.pdf](https://github.com/DEFRA/waste-exemptions-engine/files/3041080/privacy_policy.pdf)

PR #121, opened 2019-04-05
Attach privacy policy pdf to confirmation

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-208

[PR 119](https://github.com/DEFRA/waste-exemptions-engine/pull/119) created a html view which is used to generate the privacy policy pdf. This change adds the generated pdf to the project (added to lib as it seems to be the best place for it!) and amends the confirmation mailer code to read the file and add it as an attachment.


app/mailers/waste_exemptions_engine/confirmation_mailer.rb
```diff
@@ -42,5 +43,12 @@ def generate_pdf_certificate
       nil
     end
 
+    def privacy_policy
+      root_path = File.join(__dir__, "..", "..", "..")
+      privacy_pdf_path = File.absolute_path(File.join(root_path, "lib/privacy_policy.pdf"))
```
cintamani: What about using `Rails.root.join("lib/privacy_policy.pdf")`? I'm not 100% sure it will work on Rails 4 but it should :) Worth testing it :D

spec/factories/registration.rb
```diff
@@ -25,5 +25,13 @@
     trait :sole_trader do
       business_type { "soleTrader" }
     end
+
+    # This is the minimum needed to be able to generate a confirmation email
+    # and its attachments sucessfully
+    trait :emailed do
```
cintamani: If this is the minimum "required" to generate an email I find the naming not quite accurate. What do you think about de-coupling this trait's name from the purpose it has been generated for and go for a more general naming? I mean, this trait actually creates a `valid` or `submitted` or `complete` Registration object, which by default will be "emailable" if that makes sense :D

app/mailers/waste_exemptions_engine/confirmation_mailer.rb
Cruikshanks: As it's an engine and not in a rails app directly I didn't think to try it. But as this engine will only ever be used in a rails app I will!

spec/factories/registration.rb
Cruikshanks: Completely get your point, only issue is just adding these attributes doesn't make it any of those i.e.

- it's not `valid` because its missing a bunch of data
- it could be `submitted` but the definition of a submitted registration is that the record exists, so you could argue no data whatsoever is needed to meet this definition
- I think `complete` is more relevant to a transient_registration, as again we're back to the idea that a `submitted` or `complete` transient_registration is simply a registration. Also IMHO in the terms of the factory I'd assume `complete` refers to a fixture with all data populated

So I'm stuck on naming. Perhaps `emailable` is a better definition (and according to the internet it is a word no matter what my spell checker says! https://en.oxforddictionaries.com/definition/emailable) but I'm open to anything better.

cintamani: Naming things is always the difficult part! :D I think `emailable` is the best choice here :)

app/mailers/waste_exemptions_engine/confirmation_mailer.rb
Cruikshanks: ☹️ No go. `Rails.root` works, but it's not where I need to be. The pdf is in the engine (Gem's) `lib/` folder and as such `Rails.root` puts me in the wrong place.

For example in our vagrant environment when running the tests `Rails.root.join("lib/privacy_policy.pdf")` becomes `/vagrant/waste-exemptions-engine/spec/dummy/lib/privacy_policy.pdf`. I need it to be `/vagrant/waste-exemptions-engine/lib/privacy_policy.pdf`.

cintamani: Oh! Well, this is good to know for the future then :D

PR #126, opened 2019-04-12
[WIP] Add interceptors registration for action mailer in an initializer file.

cintamani:
Do not load the interceptor if the feature's config is off.
Fix last email test matching error string against error string.


cintamani:
This is actually not needed. The error appeared only when the test is run in isolation.

PR #129, opened 2019-04-16
Move confirmation email PDF logic to own service

cintamani:
Part of: https://eaflood.atlassian.net/browse/RUBY-223

This PR introduces some changes to the generation of the confirmation PDF:x
 - The `view_context` object is actually not used by the generation process, so it has been removed
 - The logic that generates the PDF itself as been moved in a Service object
 - Test coverage added in a register and re-play pattern
 - Added spec support for time travel using rails Time helpers
 - Added PDF file content matchers to spec support


app/services/waste_exemptions_engine/confirmation_pdf_generator_service.rb
```diff
@@ -0,0 +1,36 @@
...
+      )
+    end
+
+    def locals
+      {
+        presenter: presenter
+      }
+    end
+
+    def presenter
```
Cruikshanks: Same point about one line methods. Is there a need for this rather than calling `CertificatePresenter.new(@registration)` directly?

spec/factories/registration.rb
```diff
@@ -2,6 +2,12 @@
 
 FactoryBot.define do
   factory :registration, class: WasteExemptionsEngine::Registration do
+    trait :submitted do
```
Cruikshanks: As far as I can tell we don't actually use the sequenced reference in the new test added, and instead define a specific one when we call `create()`.

So unless I have missed something I would drop this until it's actually being used.

spec/services/waste_exemptions_engine/confirmation_pdf_generator_service_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+require "rails_helper"
+
+module WasteExemptionsEngine
+  RSpec.describe ConfirmationPdfGeneratorService do
+    describe ".send_confirmation_email" do
+      it "generates and return a string containing PDF content with the confirmation information" do
+        travel_to Time.local(2018, 1, 1, 1, 5, 0)
+        registration = create(:registration, :submitted, reference: "REFERENCE-1")
+        registration.addresses << create(:address, :site_address)
```
Cruikshanks: We probably haven't spoken about it (and should probably write it down somewhere 😩 😳 ) but we agreed after the Waste Carriers work to adopt a pattern of all factory object creation being driven through traits, and to avoid doing any additional changes to the object in the tests.

So ideally adding the site address should be moved to the `:submitted` trait. And if 'submitted' no longer fits the bill because of this, possibly change the name (`:confirmable`, `:confirmation_pdf`, `:i_dont_know_cause_naming_things_sucks`).

spec/support/pdf_matchers.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require "rspec/expectations"
+require "pdf-reader"
+
+RSpec::Matchers.define :match_pdf_content do |fixtures_path_to_expected_pdf|
```
Cruikshanks: Was wary of getting into trying to check the contents of PDF's, but think this is an awesome solution ❤️ 🎉 !

Just means I will now need to find time to go back and use it in our other PDF stuff 😩 .

spec/support/pdf_matchers.rb
```diff
@@ -0,0 +1,25 @@
...
+  match do |actual|
+    expected = PDF::Reader.new(fixtures_path_to_expected_pdf)
+    @expected = expected.pages.map(&:text).join
+    @expected = @expected.delete("\n").delete(" ").delete("\t")
+
+    temp_pdf = Tempfile.new("pdf")
+    temp_pdf.binmode
```
Cruikshanks: I had to go away and google what `temp_pdf.binmode` does, though having done so it now makes perfect sense. As there will be future new devs who might come across this, and it looks a bit random (`binmode.include?(:swing, :pedal, :touch)` 😁 ) could we get a brief comment on what it is and why it's needed?

spec/services/waste_exemptions_engine/confirmation_pdf_generator_service_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+module WasteExemptionsEngine
+  RSpec.describe ConfirmationPdfGeneratorService do
+    describe ".send_confirmation_email" do
+      it "generates and return a string containing PDF content with the confirmation information" do
+        travel_to Time.local(2018, 1, 1, 1, 5, 0)
```
Cruikshanks: For now this needs to be changed to use `Timecop` e.g. `Timecop.freeze(Time.local(2018, 1, 1, 1, 5, 0))`.

I expand on this in a later comment. In case you need it https://github.com/travisjeffery/timecop#usage

spec/support/time_helpers.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+RSpec.configure do |config|
```
Cruikshanks: Following on from a previous comment we are currently using [Timecop](https://github.com/travisjeffery/timecop) for any date stuff in the tests.

However what do you know, you can do this with just rspec! 🤦‍♂️ 

So I think we should move to using the rspec timehelpers, but I think it should be done as a single task rather than have a mix of ways of playing with time in the tests. So till then I'd be looking to remove this and stick to using Timecop to keep things consistent.

app/services/waste_exemptions_engine/base_service.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module WasteExemptionsEngine
```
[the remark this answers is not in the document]
cintamani: Your approach would work as well :) I was thinking that if we decide what the pattern should be for our service objects, having a base class will help maintain the convention moving forward. About the name of the method, services are not `generators`, if you know what I mean.  So we should keep it generic. Any service object shall have only one public method, usually at a class-level, and that method is usually a `call` or `run` depending on team tastes :D

app/services/waste_exemptions_engine/confirmation_pdf_generator_service.rb
```diff
@@ -0,0 +1,36 @@
...
+  class ConfirmationPdfGeneratorService < BaseService
+    def run(registration:)
+      @registration = registration
+
+      pdf_generator.pdf
+    end
+
+    private
+
+    def pdf_generator
```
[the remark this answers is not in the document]
cintamani: Sure it can. But, since the PDF generation service is an external resource to this service, is good practice to keep it in its own method. I suggest this reading: https://www.poodr.com/ it exposes very clearly the facts for such choices :)

app/services/waste_exemptions_engine/confirmation_pdf_generator_service.rb
cintamani: Ditto of https://github.com/DEFRA/waste-exemptions-engine/pull/129/files#r276214828 :D

spec/factories/registration.rb
cintamani: My mistake! I should kill the `reference` from this trait. I will keep the `submitted_at` thought, we do use it. Without the PDF generation fails as the view expects the value to be present.

spec/services/waste_exemptions_engine/confirmation_pdf_generator_service_spec.rb
cintamani: Nice! I'll address this :)

spec/support/pdf_matchers.rb
cintamani: Of course! Is mostly to deal with the way the PDF is generated and store in the different OS. I'll add a comment :D

app/services/waste_exemptions_engine/base_service.rb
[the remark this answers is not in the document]
Cruikshanks: That makes sense. So go with what you proposed for now, and lets try and have a discussion and get agreement in the next dev session.

spec/support/time_helpers.rb
cintamani: I have seen Timecop in the gemspec, but I couldn't find any single test that uses Timecop in this codebase. Are you sure it is actually used? Because if not, we can just remove the gem and keep this support file :D

cintamani:
@Cruikshanks do you mind taking a look at my answers before I start making changes? Just so I can do all the changes at once with more clarity of what should definitely change  :)

spec/support/time_helpers.rb
cintamani: On a side note, this support is not coming from RSpec but from Rails. I think it was introduced in Rails 4.2. It definitely wasn't' supported by Rails 3 :D

Cruikshanks: Think you are right, was probably me chucking it in as a copy and paste job from Waste Carriers.

However I'd still like to stick with Timecop for now, because when I talk about consistency I mean across both services. And when we switch to using the rspec helpers it would be across both as well.

I just don't like the idea of having one style in WCR, and another one in WEX. However @irisfaraway does most of the work anyways, so if she feels differently I'm happy to pipe down 😁

app/services/waste_exemptions_engine/confirmation_pdf_generator_service.rb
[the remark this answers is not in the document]
Cruikshanks: Leave as submitted for now, and let's add it as another thing we discuss and agree next Tuesday in our dev meeting. You can also give us the TL;DR as life gives me about 3 minutes a day for reading, so it'd be awhile till I get to those facts!

Cruikshanks:
@cintamani Have responded to your comments. So only changes I'm still pushing for are

- updates to the new trait in the registration factory
- moving the site creation and assignment out of the test and into the trait
- drop using active support timehelpers in favour of Timecop for now

app/services/waste_exemptions_engine/confirmation_pdf_generator_service.rb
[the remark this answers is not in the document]
cintamani: Long story short, there are 2 main benefits: 
- As the class grow (if/when) the method name is also documenting the purpose of the object or the logic inside the method. 
- As the code change, and so the PDF generator can change, having the instance created in its own method means that will be easier to update it to a different object name, given that the public interface of the generator object is unchanged.

spec/support/time_helpers.rb
cintamani: I'm happy either way too :) is an easy transition anyway as the syntax is basically the same :D So is an @irisfaraway decision ;)

cintamani: For now, I have used Timecop, will be easy to bring back the Time helper support stuff :) let's discuss it on Tuesday

PR #134, opened 2019-04-24
Update confirmation PDF to reflect the latest layout and design

cintamani:
https://eaflood.atlassian.net/browse/RUBY-223

- Added tests coverage and refactoring of factories
- Added pdf_matchers RSpec helpers to match PDF content
- Changed DatabaseCleaner strategy (still unsure what's causing failure on unique constraints - will need more investigation)
- Removed usage of seeded Exemption entries and used factories instead


spec/factories/address.rb
```diff
@@ -2,8 +2,24 @@
...
+  modes = WasteExemptionsEngine::Address.modes
 
   factory :address, class: WasteExemptionsEngine::Address do
+    trait :manual do
+      mode { modes[:manual] }
+      premises { "#{address_type} Address Premises" }
+      street_address { "#{address_type} 34 Honey road" }
+      locality { "#{address_type} Avon" }
+      city { "#{address_type} Bristol" }
+      postcode { "#{address_type} BS1 5AH" }
```
cintamani: Since the PDF contains all the addresses added in the registration phase, and I have the need to check that a particular address has been added for a specific type, this will make sure that when I check for the "site manual address" (https://github.com/DEFRA/waste-exemptions-engine/pull/134/files#diff-192d7a07fa048762e8a972439ffdd1b7R42) the test doesn't pass just because other addresses in the PDF are equal to this. Open to suggestions if this is not right :)

spec/support/pdf_matchers.rb
```diff
@@ -3,30 +3,56 @@
...
+  failure_message do |_actual|
+    "expected that \n#{@actual} \n would be equal to \n#{@expected}"
+  end
+end
 
-    temp_pdf << actual
-    @actual = PDF::Reader.new(temp_pdf).pages.map(&:text).join
-    @actual = @actual.delete("\n").delete(" ").delete("\t")
-    temp_pdf.close
+RSpec::Matchers.define :include_pdf_content do |expected|
```
cintamani: @Cruikshanks I was thinking that we should probably get rid of the `match_pdf_content` and only use this instead. Will make it clearer what information has actually been rendered in the PDF when reading the test. Here is the proposal code: https://github.com/DEFRA/waste-exemptions-engine/compare/223-tris...223-tris-matchers-proposal?expand=1 

Benefits: 
 - No need for Timecop, information can stay dynamic 
 - Allow the usage of gems like [Faker](https://github.com/stympy/faker/) that can really help to dry out the Factories 
 
The only drawback I can see is the length of the test. 
But, I actually prefer to be able to open a test and understand all that is going on. Error messages and regression fixes might also come much easier as the error will tell us which one is the information that has changed.

app/presenters/waste_exemptions_engine/certificate_presenter.rb
```diff
@@ -2,53 +2,26 @@
...
-        I18n.t("#{LOCALES_KEY}.changes.year")
-      )
+    def applicant_name
+      format("%<first>s %<last>s", first: applicant_first_name, last: applicant_last_name)
     end
 
-    def site_is_grid_reference?
-      site_address.auto?
+    def business_type
+      super.underscore.humanize
```
irisfaraway: I'm not sure we can count on `humanize` to get a readable result – the values we have in the form do diverge slightly from what is stored in the DB. Here's what we have on the form:

```
          individual: Individual or sole trader
          limited_company: Limited company
          partnership: Partnership
          limited_liability_partnership: Limited liability partnership
          local_authority: Local authority or public body
          charity: Charity or trust
```

I think it'd be good to match that on the PDF if we can.

app/presenters/waste_exemptions_engine/certificate_presenter.rb
```diff
@@ -2,53 +2,26 @@
...
+      end.join("</br>").html_safe
     end
 
-    def expires_after_pluralized
-      ActionController::Base.helpers.pluralize(
-        WasteExemptionsEngine.configuration.years_before_expiry,
-        I18n.t("#{LOCALES_KEY}.changes.year")
-      )
+    def applicant_name
+      format("%<first>s %<last>s", first: applicant_first_name, last: applicant_last_name)
```
irisfaraway: Can we do any refactoring to reduce the duplicated `format("%<first>s %<last>s", first: x_first_name, last: x_last_name)`, since all 3 uses are very similar aside from the source of the names?

app/views/waste_exemptions_engine/pdfs/certificate.html.erb
```diff
@@ -1,192 +1,189 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <meta charset="utf-8">
-    <title></title>
+    <meta charset="utf-8" />
```
irisfaraway:
```suggestion
    <meta charset="utf-8">
```

I think we can ditch ending `/`s in void tags in HTML5 👍

app/views/waste_exemptions_engine/pdfs/certificate.html.erb
```diff
@@ -1,192 +1,189 @@
...
+            </p>
 
-        <!-- Header section -->
-        <header class="page-header group">
-          <div class="hgroup">
-            <h1 class="heading-small"><%=t ".heading" %></h1>
-          </div>
-        </header>
+            <p class="govuk-body">
+              <strong> <%=t ".confirmation.warning" %> </strong>
```
irisfaraway:
```suggestion
              <strong><%=t ".confirmation.warning" %></strong>
```

Maybe a nitpick, but can we remove this extra spacing? There are a few other tags with this spacing on the page (mostly `td`s) but it doesn't look consistent.

config/locales/pdfs/certificate.en.yml
```diff
@@ -2,41 +2,64 @@ en:
...
+        confirmation:
+          heading: Waste exemption registration %{registration_number}
+          paragraph1: We confirm that the details you have entered, as shown below, are now on the exemptions register. Your registration should show on the public register within 5 days.
+          warning: This document is not a permit. Exemptions may have ceased or been revoked.
+          paragraph2: "To check whether these exemptions are still valid, check the company’s registration %{company_registration_number} on the public register of waste exemptions:"
+        business_responsibilities:
+          heading: Business responsibilities
+          paragraph1_html: >
+            The business or organisation responsible for carrying out the exempt waste operations agrees to:
+            <ul class="govuk-list govuk-list--bullet govuk-!-margin-left-2">
```
irisfaraway: I'd prefer to break this list into individual lines of YAML and keep all the styling in the HTML template – this is more consistent with how we've done other lists the past (example: https://github.com/DEFRA/waste-exemptions-engine/blob/master/config/locales/mailers/confirmation_mailer.en.yml#L12-L15) We've generally tried to avoid including HTML in the YAML unless it's something really tiny like a `<strong>` tag.

spec/factories/registration.rb
```diff
@@ -20,23 +20,44 @@
...
+
+      contact_first_name { "Contact First Name" }
+      contact_last_name { "Contact Last Name" }
+      contact_phone { "07459745569" }
+      contact_email { "contact@test.defra.gov.uk" }
+      contact_position { "Manager" }
+
+      addresses { [build(:address, :site_address, :auto), build(:address, :operator_address, :manual), build(:address, :contact_address, :manual)] }
+
+      after(:build) do |registration, _|
```
irisfaraway: I'm just being nosy here :) What's the `_` argument doing?

spec/forms/waste_exemptions_engine/exemptions_form_spec.rb
```diff
@@ -12,7 +12,7 @@ module WasteExemptionsEngine
     end
 
     subject(:form) { build(:exemptions_form) }
-    let(:three_exemptions) { Exemption.order("RANDOM()").last(3) }
+    let(:three_exemptions) { create_list(:exemption, 3) }
```
irisfaraway: Now that we are using `create_list` here instead of Exemptions from the DB, is the comment in the `before` block above still accurate?

spec/requests/waste_exemptions_engine/exemptions_forms_spec.rb
```diff
@@ -9,7 +9,7 @@ module WasteExemptionsEngine
     include_examples "GET form", :exemptions_form, "/exemptions"
     include_examples "go back", :exemptions_form, "/exemptions/back"
     include_examples "POST form", :exemptions_form, "/exemptions" do
-      let(:form_data) { { exemptions: Exemption.all.map(&:id).map(&:to_s) } }
+      let(:form_data) { { exemptions: create_list(:exemption, 10).map(&:id).map(&:to_s) } }
```
irisfaraway: Also wondering if the `before(:context) { create_list(:exemption, 5) }` on line 7 is affected or made redundant by this change.

spec/support/database_cleaner.rb
```diff
@@ -7,10 +7,13 @@
   # Clean the database before running tests. Setup as per
   # https://github.com/DatabaseCleaner/database_cleaner#rspec-example
   config.before(:suite) do
-    DatabaseCleaner.strategy = :transaction
     DatabaseCleaner.clean_with(:truncation)
   end
 
+  config.before(:each) do
+    DatabaseCleaner.strategy = :truncation
```
irisfaraway: Is this fixing the issue with the unique keys? The commit message sounded like we weren't sure if it had solved the problem or not.

app/presenters/waste_exemptions_engine/certificate_presenter.rb
cintamani: Oh nice one! :D

config/locales/pdfs/certificate.en.yml
cintamani: will do :)

spec/factories/registration.rb
cintamani: just convention to say "there is a param there, but I'm not going to use it" :D

spec/forms/waste_exemptions_engine/exemptions_form_spec.rb
cintamani: Oh that's were the DB was populated from! 😂 missed it. I think the entire `before` block can go. Thanks for pointing that out :)

spec/support/database_cleaner.rb
cintamani: It does solve it yes :)

spec/factories/registration.rb
irisfaraway: Ahh OK, I've seen that with a param name afterwards (eg `_foo`) but never just on its own. What is the second param that we are ignoring? The other `after` blocks in our factories only seem to take one param, the object created by the factory.

cintamani: Well, I assume that if you define only one the second just gets lost. I'll remove it is not necessary :) (it is the `evaluator` the object passed there)

cintamani:
Closing this as a new PR will land for this change later with the changes agreed with Alan

PR #136, opened 2019-04-24
Remove byebug and swap it with pry-byebug

cintamani:
byebug has a very similar interface as gdb, but byebug does not use the powerful Pry REPL.
binding.pry uses Pry, but lacks some of the byebug features.
[pry-byebug](https://github.com/deivid-rodriguez/pry-byebug) brings some capabilities byebug to binding.pry, so using that, will give you the most debugging powers.

Credits: https://docs.gitlab.com/ee/development/pry_debugging.html#byebug-vs-bindingpry

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -66,6 +66,7 @@ GEM
       activesupport (>= 3.0.0)
       uniform_notifier (~> 1.11)
     byebug (11.0.1)
+    coderay (1.1.2)
     concurrent-ruby (1.1.5)
     crack (0.4.3)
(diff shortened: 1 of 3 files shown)


cintamani:
Sure :) I'll add more details

cintamani:
@Cruikshanks I was actually thinking that we might get all-the-power if we use https://github.com/deivid-rodriguez/pry-byebug WDYT?

Cruikshanks:
Go for it @cintamani !

cintamani:
@Cruikshanks ready for review :)

PR #138, opened 2019-04-24
Subclass TransientRegistration with NewRegistration

irisfaraway:

This is part of a series of PRs to add editing functionality to the back office.

When we first created TransientRegistration, it was only used for the new registration journey. However, now we want to have multiple times of TransientRegistrations for different occasions when we want to modify Registration data.

This PR adds the ability to subclass TransientRegistration, and also creates a NewRegistration to perform the functions specific to the new registration journey.


irisfaraway:
File count is high, unfortunately – however the vast majority of this is just from swapping "TransientRegistration" to "NewRegistration" in tests.

irisfaraway:
This has some implications for the back office, which are dealt with here: https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/141

The front office will need a database migration to add `type` to the TransientRegistration schema, but that's about it.

app/models/waste_exemptions_engine/new_registration.rb
```diff
@@ -0,0 +1,16 @@
...
+module WasteExemptionsEngine
+  class NewRegistration < TransientRegistration
+    include CanUseNewRegistrationWorkflow
+
+    after_create :apply_reference
+
+    private
+
+    def apply_reference
+      self.reference = format("WEX%06d", id)
```
cintamani: TIL! I didn't know about this `format` method! 😆 nice!

app/models/waste_exemptions_engine/transient_registration.rb
```diff
@@ -37,15 +34,8 @@ def registration_attributes
           "temp_grid_reference", "temp_site_description",
           "address_finder_error", "transient_addresses",
           "transient_registration_exemptions", "transient_people",
-          "created_at", "updated_at"
+          "created_at", "updated_at", "type"
```
cintamani: Wow! This is a quite long list of attributes! :D Would the other way around generate less code (`take` the ones we need rather than `except` the one we don't)? Might also be the case to externalise this list in some constant for better visibility and easier changes

db/migrate/20190424143344_add_type_to_transient_registrations.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddTypeToTransientRegistrations < ActiveRecord::Migration
+  def change
+    add_column :transient_registrations, :type, :string
```
cintamani: Instead of creating a different migration, what about:
```
def up 
  add_column :transient_registrations, :type, :string
  execute <<-SQL
     UPDATE transient_registrations 
     SET type = 'TransientRegistration';
  SQL
end

def down
  remove_column ....
end
```

spec/dummy/db/schema.rb
```diff
@@ -43,6 +43,12 @@
 
   add_index "addresses", ["registration_id"], name: "index_addresses_on_registration_id", using: :btree
 
+  create_table "defra_ruby_exporters_bulk_export_files", force: :cascade do |t|
+    t.string   "file_name"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
```
cintamani: I don't think this should be part of this PR :)

spec/dummy/db/schema.rb
```diff
@@ -183,6 +189,7 @@
     t.boolean  "address_finder_error",   default: false
     t.datetime "created_at",                             null: false
     t.datetime "updated_at",                             null: false
+    t.string   "type"
```
cintamani: Shall we add some validations to this? Also a `null: false` feels like is needed. This looks like 1 step in the Single Table Inheritance direction, hence the `type` attribute is going to play a fundamental role going forward :)

spec/dummy/db/schema.rb
irisfaraway: I was being lazy but you are right, should be in a separate one. Removed in https://github.com/DEFRA/waste-exemptions-engine/pull/138/commits/a0bc400746aecc57ff0119154702293a08ca3836

spec/dummy/db/schema.rb
irisfaraway: Added a `null: false` in https://github.com/DEFRA/waste-exemptions-engine/pull/138/commits/34ecbe11f8c572198c348a054ea472ad4e5a26bc

app/models/waste_exemptions_engine/transient_registration.rb
irisfaraway: Good idea – moved to a constant in https://github.com/DEFRA/waste-exemptions-engine/pull/138/commits/6ad41c6c1b5620cd14c1376afc15a2036686bcaa

I think the list of attributes to `take` is slightly longer than the `except` ones, so have left it this way around for now.

spec/dummy/db/schema.rb
irisfaraway: Oops, broke everything. Ignore that for now...

db/migrate/20190424143344_add_type_to_transient_registrations.rb
irisfaraway: Implemented! https://github.com/DEFRA/waste-exemptions-engine/pull/138/commits/946e1f81cef9b7e93fd1463503a3321a044e2025

spec/dummy/db/schema.rb
irisfaraway: Required a few other commits first to refactor, but think I've sorted this now: https://github.com/DEFRA/waste-exemptions-engine/pull/138/commits/938be38d708a451ee4f956808e2dfeb6367a081d

app/models/waste_exemptions_engine/transient_registration.rb
```diff
@@ -28,24 +25,26 @@ class TransientRegistration < ActiveRecord::Base
...
-          "id", "token", "workflow_state", "start_option",
-          "declaration", "temp_operator_postcode",
-          "temp_contact_postcode", "temp_site_postcode",
-          "temp_grid_reference", "temp_site_description",
-          "address_finder_error", "transient_addresses",
-          "transient_registration_exemptions", "transient_people",
-          "created_at", "updated_at"
-        )
-    end
+    ATTRIBUTES_NOT_COPIED_TO_REGISTRATION = %w[address_finder_error
```
cintamani: What about calling these `TRANSIENT_ATTRIBUTES`? In general, I find a good practice to name something against what it *is* rather than against what it *isn't* :)

irisfaraway:
@cintamani https://github.com/DEFRA/waste-exemptions-engine/pull/139 will add a new subclass of TransientRegistration specifically for editing.

Depending on how we implement renewals, we may want a third subclass for that down the line as well. I definitely think we will need all 3 types in WCR.

app/models/waste_exemptions_engine/transient_registration.rb
irisfaraway: Yeah, that's also a lot shorter 😅 I'll update it!

irisfaraway: Changed in https://github.com/DEFRA/waste-exemptions-engine/pull/138/commits/a3e4a617837f2c228d71e0e54c306ec9e4884ff0

irisfaraway: @cintamani Sorry, it voided your review!

PR #142, opened 2019-04-26
Cancel an edit

irisfaraway:

It should be possible for users to cancel an in-progress edit.

This should destroy the EditRegistration without saving any of the modified data to the original Registration.


irisfaraway:
CodeClimate has a gripe about the number of methods in the EditFormsController. Unfortunately this is down to the number of different state machines transitions possible from that state. I'm not sure how we can modify this but would appreciate ideas for refactoring (if anyone thinks it is worth it).

app/views/waste_exemptions_engine/edit_cancelled_forms/new.html.erb
```diff
@@ -0,0 +1,26 @@
+
```
cintamani: [NIT]
```suggestion
```

app/views/waste_exemptions_engine/edit_cancelled_forms/new.html.erb
```diff
@@ -0,0 +1,26 @@
...
+  <h1 class="heading-large"><%= t(".error_heading") %></h1>
+
+  <% @edit_cancelled_form.errors.full_messages.each do |message| %>
+    <li><%= message %></li>
+  <% end %>
+
+<% else %>
+
+  <%= form_for(@edit_cancelled_form) do |f| %>
+    <%= render("waste_exemptions_engine/shared/errors", object: @edit_cancelled_form) %>
```
cintamani:
```suggestion
```

irisfaraway: This had the same issues as the EditCompleteForm from the other PR. So I've done a matching fix: https://github.com/DEFRA/waste-exemptions-engine/pull/142/commits/4c00561259db806438981d2cec23a3b222b90268

irisfaraway:
@cintamani Agree it was a pretty awkward name – I tried a few things but was getting hung up on starting it with "Edit". Have gone with "ConfirmEditCancelledForm" instead.

irisfaraway:
Actually, hold off on reviewing this for now – need to check the styling.

irisfaraway:
Styling added (although it won't be visible until https://github.com/DEFRA/waste-exemptions-front-office/pull/102 and https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/152 are merged) – ready for a review!

PR #147, opened 2019-04-29
Add a feature switch and permissions check for editing

irisfaraway:

Right now we only want to enable editing in the back office. So we need a switch to enable or disable the feature. Host apps can set this in the engine initialiser.

This PR also introduces a permissions check for editing. This is set to always pass by default in the engine but is intended to be overwritten in the host apps when needed.

---

For an example of these host app changes, see https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/153


app/controllers/waste_exemptions_engine/forms_controller.rb
```diff
@@ -112,6 +112,18 @@ def form_matches_state?
       controller_name == "#{@transient_registration.workflow_state}s"
     end
 
+    def edit_checks_pass?
+      return true unless @transient_registration.is_a?(EditRegistration)
+      return true if edit_enabled? && current_user_can_edit?
+
+      redirect_to "/errors/404", status: 404
```
cintamani:
```suggestion
      redirect_to errors_path(id: 404), status: 404
```

app/controllers/concerns/waste_exemptions_engine/permission_checks.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# This concern is responsible for performing permissions checks. It's intended
+# to be included overriden in host applications which have their own specific
+# permissions and roles.
+module WasteExemptionsEngine
+  module PermissionChecks
+    def current_user_can_edit?
+      true
```
cintamani: In these cases I used to: 
```
raise(NotImplementedError) 
```

app/controllers/waste_exemptions_engine/forms_controller.rb
cintamani:
```suggestion
      redirect_to errors_path(id: 404), status: 404
```

app/controllers/concerns/waste_exemptions_engine/permission_checks.rb
irisfaraway: I've done this in the past but was thinking that it was better to leave `true` in as a default - do you think it's better to force host apps to implement their own version of the method? I was thinking it'd be harder to test and manage, but also I can see the benefits of forcing apps to define all their own permissions, although the front-office doesn't really have any concept of them yet...

cintamani: I am concerned about the fact that we might give unauthorised people the ability to change exemptions mostly. What if we raise the exception only in the environments that have the feature turned on then?

irisfaraway:
Oops, path change has broken the test – will fix and then re-request a review!

app/controllers/concerns/waste_exemptions_engine/permission_checks.rb
irisfaraway: Editing is currently disabled in the front office so it doesn't feel like a pressing issue – however if we raise a `NotImplementedError` here it will also make our tests more complicated as the dummy app will have to have its own EditHelper config, or we'll have to do a lot more stubbing in the tests themselves. So it feels like more pain than benefit for me at the moment.

PR #148, opened 2019-04-30
Raise error in completion service

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-221

In order to test RUBY-221 we need to raise an error in the completion service to ensure that the user sees an error page rather than the completion page.

DO NOT MERGE IN!


Cruikshanks:
Should never have raised the PR, just a force of habit!

PR #150, opened 2019-04-30
Update confirmation PDF to reflect the latest layout and design

cintamani:
https://eaflood.atlassian.net/browse/RUBY-223

Change the style and layout of confirmation PDF as per ticket's design direction.
Added gov.uk css style and used to style the document in line with the general website style.
The document is created dynamically and render information in an ad-hoc format depending on the nature
of the business and the site addresses types.
<details><summary>An example of document generated by this new code</summary>
<p>

Screenshot 2019-04-30 at 16 14 53
Screenshot 2019-04-30 at 16 15 00

</p>
</details>


spec/factories/address.rb
```diff
@@ -2,6 +2,7 @@
 
 FactoryBot.define do
   address_types = WasteExemptionsEngine::Address.address_types
+  modes = WasteExemptionsEngine::Address.modes
```
Cruikshanks: ❤️

spec/services/waste_exemptions_engine/confirmation_pdf_generator_service_spec.rb
```diff
@@ -4,18 +4,71 @@
...
+        expect(result).to include_pdf_content(registration.operator_name)
+        expect(result).to include_pdf_content(registration.contact_phone)
+        expect(result).to include_pdf_content(registration.contact_email)
+        expect(result).to include_pdf_content(registration.contact_first_name)
+        expect(result).to include_pdf_content(registration.contact_last_name)
+        expect(result).to include_pdf_content(registration.contact_position)
 
-        expect(result).to match_pdf_content(fixtures_file_path)
+        registration.addresses.each do |address|
+          if address.manual?
```
Cruikshanks: It's not a great name 😩 but if the address mode is `auto` then we know we're dealing with the site address in its grid reference and description configuration.

So I think it would be more accurate to say

```ruby
  if address.auto?
    expect(result).to include_pdf_content(address.description)
    expect(result).to include_pdf_content(address.grid_reference)
  else
    expect(result).to include_pdf_content(address.premises)
    expect(result).to include_pdf_content(address.street_address)
    expect(result).to include_pdf_content(address.locality)
    expect(result).to include_pdf_content(address.city)
    expect(result).to include_pdf_content(address.postcode)
  end
```

spec/support/pdf_matchers.rb
```diff
@@ -3,30 +3,41 @@
 require "rspec/expectations"
 require "pdf-reader"
 
-RSpec::Matchers.define :match_pdf_content do |fixtures_path_to_expected_pdf|
-  match do |actual|
-    expected = PDF::Reader.new(fixtures_path_to_expected_pdf)
-    @expected = expected.pages.map(&:text).join
-    @expected = @expected.delete("\n").delete(" ").delete("\t")
+module Helpers
```
Cruikshanks: Assuming this will all 💥 when the other PR gets merged in and this is rebased/master is merged in!

spec/services/waste_exemptions_engine/confirmation_pdf_generator_service_spec.rb
cintamani: Make sense! What if we create a `grid_reference` method in the `address` class which for now will just check that the address mode is `auto`? Will make it much clearer IMO

Cruikshanks: I get what you mean and 100% agree. I'm just not sure of the name as we have a grid_reference attribute. So this could mean return me that, or has grid reference has been set or not.

So none of these are any better but `uses_grid_reference?`, `located_by_grid_reference?`, `address_is_a_grid_reference?` I think give a bit more indication of the intent/purpose of the method.

I just hope you can do better than my attempts!

Cruikshanks:
I swear I had other comments (else my overall review comment doesn't make sense!) 😖 😕 

Will try and add them again now 😩

app/views/waste_exemptions_engine/pdfs/certificate.html.erb
```diff
@@ -1,192 +1,199 @@
...
+                  <% if presenter.partnership? %>
+                    <tr class="govuk-table__row">
+                      <th class="govuk-table__header" scope="row">
+                        <%=t ".registration_details.business_details.partners" %>
+                      </th>
+                      <td class="govuk-table__cell"> <%= presenter.partners_names %> </td>
+                    </tr>
+                  <% else %>
+                    <tr class="govuk-table__row">
+                      <th class="govuk-table__header" scope="row"><%=t ".registration_details.business_details.responsible_person" %></th>
```
Cruikshanks: I normally try and make a clear distinction between code review and QA. So I'll focus on quality and consistency in the review, and whether it actually does what it's supposed to in QA.

From a code perspective this is fine but I can (as long as I have understood the code correctly 😁 ) immediately see it will fail QA.

The labels are still referring to **Person responsible ...**. However this will only make sense in the context of a sole trader. If the registrant is Waitrose this will read as

> **Person responsible for these waste operations**: Waitrose Ltd

So as this is still not working for all contexts and clearly so, I don't there is any point approving and merging until we have rectified this issue.

cintamani: nice! Is this knowledge of all cases you have that I am missing! Let's discuss it further when we got a chance :)

PR #152, opened 2019-05-01
Don't modify status of unedited exemptions

irisfaraway:

When a registration is edited, if it keeps the same exemptions, the status of those exemptions should not change.

This fixes a bug which always reset the status to active.

### DIFF
--- a/app/services/waste_exemptions_engine/edit_completion_service.rb
+++ b/app/services/waste_exemptions_engine/edit_completion_service.rb
@@ -43,10 +43,10 @@ module WasteExemptionsEngine
     end
 
     def copy_exemptions
-      @registration.exemptions = []
-      @edit_registration.exemptions.each do |exemption|
-        @registration.exemptions << exemption
-      end
+      unchanged_exemptions = @registration.exemptions & @edit_registration.exemptions
+
+      remove_unwanted_exemptions(unchanged_exemptions)
+      add_new_exemptions(unchanged_exemptions)
     end
 
     def copy_people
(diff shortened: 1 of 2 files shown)


app/services/waste_exemptions_engine/edit_completion_service.rb
```diff
@@ -43,10 +43,10 @@ def copy_addresses
...
 
     def copy_exemptions
-      @registration.exemptions = []
-      @edit_registration.exemptions.each do |exemption|
-        @registration.exemptions << exemption
-      end
+      unchanged_exemptions = @registration.exemptions & @edit_registration.exemptions
+
+      remove_unwanted_exemptions(unchanged_exemptions)
+      add_new_exemptions(unchanged_exemptions)
```
cintamani: I am pretty sure that Rails is able to deal with this itself, and will actually perform better since will make less commits to the database :) You can try with simply 
```
def copy_exemptions
  @registration.exemptions = @edit_registration.exemptions
end
```

irisfaraway: You are correct! I was worried this would ignore the data on the registration_exemption but it works fine. Clearly I was overthinking :) Fixed in https://github.com/DEFRA/waste-exemptions-engine/pull/152/commits/5017999cdf86b3829f6779e9419a5f264cbc14a3

cintamani: eheh I am pretty sure that this behaviour on Rails wasn't working correctly on previous versions so you were correct on your first assumption. Now though it behaves as expected :)

PR #156, opened 2019-05-03
Make the completition form not flexibly navigatable

cintamani:
https://eaflood.atlassian.net/browse/RUBY-312

Add flag for `can_navigate_flexibly` to false on the competition form.
This assures that anyone trying to request the page more than once will be redirected to the form start.


Cruikshanks:
A yes to both of @irisfaraway comments

app/forms/waste_exemptions_engine/registration_complete_form.rb
```diff
@@ -18,5 +18,9 @@ def initialize(registration)
     def submit(_params)
       raise UnsubmittableForm
     end
+
+    def self.can_navigate_flexibly?
```
Cruikshanks: I think the intent was to add this to the base form and default to `true` but then override it where we need to, for example here and on the declaration form.

irisfaraway: @Cruikshanks That's already in the BaseForm! I thought it had been removed but looks like not.

Cruikshanks: And that's why I set this as a comment 🤦‍♂ Thx @irisfaraway and ignore me @cintamani

cintamani: https://github.com/DEFRA/waste-exemptions-engine/pull/93 :D

cintamani:
@irisfaraway nice stuff! I'll deal with the first point in a separate PR. Can I ask you to link me to a similar shared example I can copy the style from for the second point? :)

irisfaraway:
@cintamani Have a look at https://github.com/DEFRA/waste-exemptions-engine/blob/master/spec/support/shared_examples/requests/get_form.rb#L49-L56

I think the declaration form request spec will already be using this shared example, so you shouldn't need to modify the tests for that. The "registration complete" page doesn't use it because it's got its own special stuff going on with the transient_registration being deleted and all that.

If we really want to go in hard on the tests, we could add a test to the form object checking its response to `self.can_navigate_flexibly?`

cintamani:
Yep I wanted to add that unit test too. I will check the shared example :) Thanks for the help <3

cintamani:
@Cruikshanks I don't have permissions to merge this without that codeclimate check being green. Can you push the button for me?

PR #157, opened 2019-05-03
Refactoring of factories to use the Faker gem

cintamani:
As part of our Development style discussion we agreed that Faker is a good choice to introduce in our world of factories.
This code introduce the gem and refactor the tests to use it when appropriate.


spec/factories/registration.rb
```diff
@@ -38,39 +38,40 @@
     # This is the minimum needed to be able to generate a confirmation email
     # and its attachments sucessfully
     trait :emailable do
-      reference { "WEX000999" }
```
cintamani: Removed since a reference is defined in the main factory.

spec/services/waste_exemptions_engine/registration_completion_service_spec.rb
```diff
@@ -57,10 +57,10 @@ module WasteExemptionsEngine
           expect(NewRegistration.where(reference: new_registration.reference).count).to eq(0)
         end
 
-        it "sends a confirmation email" do
+        it "sends a confirmation email to both the applicant and the contant emails" do
```
cintamani: This is because now the applicant and contact emails are the same hence 2 emails get sent. I'm not sure if it is relevant / useful to create an extra scenario to test that when the emails are the same only one is sent over. Open to suggestions :D

spec/factories/new_registration.rb
```diff
@@ -27,28 +27,29 @@
...
+
       location { "england" }
-      applicant_first_name { "Joe" }
-      applicant_last_name { "Bloggs" }
+      applicant_first_name { Faker::Name.first_name }
+      applicant_last_name { Faker::Name.last_name }
       applicant_phone { "01234567890" }
-      applicant_email { "test@example.com" }
-      business_type { "limitedCompany" }
+      applicant_email { Faker::Internet.email }
```
Cruikshanks: Just to be extra, extra careful when we have used faker to generate emails in the past, we have used `Faker::Internet.safe_email` as it will generate **[name]@example.[org|com|net]**.

This is just to ensure we don't inadvertently generate a valid email address and then mistakenly send it.

```suggestion
      applicant_email { Faker::Internet.safe_email }
```

spec/factories/new_registration.rb
```diff
@@ -27,28 +27,29 @@
...
-      contact_first_name { "Joe" }
-      contact_last_name { "Bloggs" }
-      contact_position { "Chief Waste Carrier" }
+      operator_name { Faker::Company.name }
+      contact_first_name { Faker::Name.first_name }
+      contact_last_name { Faker::Name.last_name }
+      contact_position { Faker::Company.profession }
       contact_phone { "01234567890" }
-      contact_email { "test@example.com" }
+      contact_email { Faker::Internet.email }
```
Cruikshanks:
```suggestion
      contact_email { Faker::Internet.safe_email }
```

spec/factories/registration.rb
```diff
@@ -38,39 +38,40 @@
...
 
     trait :complete do
       location { "england" }
-      applicant_first_name { "Joe" }
-      applicant_last_name { "Bloggs" }
+      applicant_first_name { Faker::Name.first_name }
+      applicant_last_name { Faker::Name.last_name }
       applicant_phone { "01234567890" }
-      applicant_email { "test@example.com" }
+      applicant_email { Faker::Internet.email }
```
Cruikshanks:
```suggestion
      applicant_email { Faker::Internet.safe_email }
```

spec/factories/registration.rb
```diff
@@ -38,39 +38,40 @@
...
-      contact_first_name { "Joe" }
-      contact_last_name { "Bloggs" }
-      contact_position { "Chief Waste Carrier" }
+      operator_name { Faker::Company.name }
+      contact_first_name { Faker::Name.first_name }
+      contact_last_name { Faker::Name.last_name }
+      contact_position { Faker::Company.profession }
       contact_phone { "01234567890" }
-      contact_email { "test@example.com" }
+      contact_email { Faker::Internet.email }
```
Cruikshanks:
```suggestion
      contact_email { Faker::Internet.safe_email }
```

spec/services/waste_exemptions_engine/registration_completion_service_spec.rb
Cruikshanks: I think we actually should add something. We are using the tests to both check and document behaviour so we should add something that covers the fact that the app only sends 1 email if both applicant and contact email are the same, but 2 if they are different.

However I'm fine that this can be done later in a separate PR.

spec/support/factory_bot.rb
```diff
@@ -2,6 +2,8 @@
 
 # Require this here to get factories to play nice with engine
 require "factory_bot_rails"
+# https://github.com/stympy/faker
```
Cruikshanks: I think to make our tests more meaningful we should set the locale of Faker (`Faker::Config.locale = "en-GB"`). So as we're both requiring it and setting some config we could do with creating a `faker.rb` file in the `spec/support/` folder

```ruby
# frozen_string_literal: true

# We use this to generate fake data. We believe it is more realistic to use
# nondeterministic data in tests as it better represents actual usage and is more
# likely to highlight issues in our app
require "faker"

Faker::Config.locale = "en-GB"
```

spec/services/waste_exemptions_engine/registration_completion_service_spec.rb
cintamani: I'm happy to add it here, is related code after all :)

spec/factories/new_registration.rb
cintamani: Nice one! 👍

spec/factories/exemption.rb
```diff
@@ -8,7 +8,7 @@
     end
 
     category { WasteExemptionsEngine::Exemption.categories[:using_waste] }
-    url { "https://www.gov.uk/guidance/waste-exemptions-#{code}-using-waste" }
-    summary { "Use of waste in construction" }
+    url { "https://www.gov.uk/guidance/waste-exemptions-#{code}using--waste" }
```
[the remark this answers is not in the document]
cintamani: nope 🤔

PR #159, opened 2019-05-08
Fix error on data address lookup error handling

cintamani:
Closes https://eaflood.atlassian.net/browse/RUBY-50
Closes https://eaflood.atlassian.net/browse/RUBY-297

Since the result of AddressFinderService in case of error is a simple `symbol`, this code adds a check for the type of the service response and skips the coordinates extrapolation code in case the result is a failure.

I wanted to do a much better refactoring, by making the response become an object itself which can respond to methods like "#failed?" but unfortunately this is currently used in a lot of places and the refactoring would take out too much time. So given the amount of work needed to the release, I have parked it in my "TODO" list for quieter periods.


cintamani:
Will do that :)

PR #167, opened 2019-05-13
Move edit authorization checking to its own service

cintamani:
Part of https://eaflood.atlassian.net/browse/RUBY-330

This moves the permissions checkers in its own service to allow easier re-define of class on the engine apps

### DIFF
--- a/app/services/waste_exemptions_engine/edit_permission_checker_service.rb
+++ b/app/services/waste_exemptions_engine/edit_permission_checker_service.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+# This concern is responsible for performing permissions checks. It's intended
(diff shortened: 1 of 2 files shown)


app/services/waste_exemptions_engine/edit_permission_checker_service.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# This concern is responsible for performing permissions checks. It's intended
+# to be included overriden in host applications which have their own specific
+# permissions and roles.
+module WasteExemptionsEngine
+  class EditPermissionCheckerService < BaseService
+    def run(current_user:)
```
cintamani:
```suggestion
    def run(*)
```

PR #175, opened 2019-05-15
Fix conditionals on main people page

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-347

This PR fixes a few failures in the conditional logic used to display HTML.

It fixes the number of `div` closing tags that are used, so they should now all be closed properly, preventing the footer from running into the `main` tag in IE11.

It also makes sure that only one `ul` tag is used for the list of added people, rather than one `ul` tag per person.


irisfaraway:
Not sure if the HTML comments will be controversial, but I thought it was pretty confusing without them.

PR #177, opened 2019-05-15
Remove copy over of exemptions informations from registration to transient registration for editing

cintamani:
Following on https://github.com/DEFRA/waste-exemptions-engine/pull/171 

This removes any ability to copy over exemptions data from a completed registration to a transient object so that no editing is allowed.


app/models/waste_exemptions_engine/edit_registration.rb
```diff
@@ -6,11 +6,13 @@ class EditRegistration < TransientRegistration
 
     after_initialize :copy_data_from_registration, if: :new_record?
 
+    def registration
```
irisfaraway: We really should have added this a long time ago (especially in WCR). Love it!

spec/services/waste_exemptions_engine/edit_completion_service_spec.rb
```diff
@@ -60,39 +57,37 @@ module WasteExemptionsEngine
...
-        exemption = registration.exemptions.first
-
-        service
-        expect(registration.reload.exemptions).to_not include(exemption)
+        expect { run_service }.to_not change { registration.reload.registration_exemptions.first.state }
       end
 
-      it "adds newly-added registration_exemptions" do
-        exemption = edit_registration.exemptions.first
+      it "do not removes inactive registrations" do
```
irisfaraway:
```suggestion
      it "does not remove inactive registrations" do
```

spec/services/waste_exemptions_engine/edit_completion_service_spec.rb
```diff
@@ -60,39 +57,37 @@ module WasteExemptionsEngine
         registration.registration_exemptions.first.update(state: "foo")
         edit_registration.exemptions << registration.exemptions.first
```
irisfaraway: I was wondering why this test didn't catch the original issue with inactive exemptions? I guess it's because we are copying the new state to the `edit_registration` so it's not an accurate replication of what happens. Either way I wonder if we should change it now?

cintamani: This is because the original registration and the copy registration always loaded different exemptions. The registration is using its `complete` trait which initializes it with 10 exemptions. Then the edit was initialized with its `edited` trait, which assigned 5 completely different exemptions to the edit_registration :) (we have made subsequent changes which created this, the exemptions of the `competed` registration, for example, were merged after the edit functionality I believe :D)

irisfaraway: Cool, I think that's what I thought was going on. I was wondering if we should just get rid of the test now as the line where we copy the exemption to the `edit_registration` no longer reflects what actually happens – it's not an active status so it would not get copied over in the first place. So it seems a bit pointless now to me. What do you think?

cintamani: Good point!! I'll go with the removal of exemptions copy over :) sounds the most reasonable solutions :) We can always re-implement it when / if we decide that also exemptions can be edited :)

cintamani:
nice catch :/ What if we get the list through `edit_registration.registration.active_exemptions`?

irisfaraway:
For the other uneditable values like `location` and `business_type`, we copy them to the `edit_registration` object but just don't modify them. So I'd lean towards imitating that behaviour and keeping the `copy_exemptions_from_registration` method when we initialize. I guess it's more wasteful in that we are creating additional DB records, but it does feel more consistent with how we deal with the other uneditable data.

cintamani:
cool, will do that :)

PR #180, opened 2019-05-17
Remove duplicate partnership? method

Cruikshanks:
Spotted in a the `CertificatePresenter` that we had a `partnership?`method  however we already have that built into the registration via the `CanHaveRegistrationAttributes`.

As the `CertificatePresenter` delegates any calls it does not recognise down to the registration instance, deleting the method will have no impact on anything that uses the presenter.


Cruikshanks:
It should be noted I approved the PR that introduced this 🤦‍♂ !

PR #188, opened 2019-05-21
Do not save un-persisted objects in workflow setup

cintamani:
https://eaflood.atlassian.net/browse/RUBY-352

The registration was saved at every refresh because of the workflow setup.
Since the object is persisted already at every `submit`, I have added a guard clause to avoid updating it unless it is persisted. 
BUT the editing journey is instead expecting a persisted object since the first call. Hence a fix there was necessary too.  

I wasn't able to find a nice way to test it :/ we should run some tests in truncation mode or do some weird stubbing and expectations.

### DIFF
--- a/app/controllers/waste_exemptions_engine/forms_controller.rb
+++ b/app/controllers/waste_exemptions_engine/forms_controller.rb
@@ -79,9 +79,9 @@ module WasteExemptionsEngine
       return unless state_can_navigate_flexibly?(@transient_registration.workflow_state)
       return unless state_can_navigate_flexibly?(requested_state)
 
-      if @transient_registration.persisted?
-        @transient_registration.update_attributes(workflow_state: requested_state)
-      end
+      return unless @transient_registration.persisted?
+      
+      @transient_registration.update_attributes(workflow_state: requested_state)
     end
 
     def state_can_navigate_flexibly?(state)
--- a/app/controllers/waste_exemptions_engine/forms_controller.rb
+++ b/app/controllers/waste_exemptions_engine/forms_controller.rb
@@ -78,9 +78,20 @@ module WasteExemptionsEngine
     def set_workflow_state
       return unless state_can_navigate_flexibly?(@transient_registration.workflow_state)
       return unless state_can_navigate_flexibly?(requested_state)
-      return unless @transient_registration.persisted?
 
-      @transient_registration.update_attributes(workflow_state: requested_state)
+      if @transient_registration.persisted?
+        update_workflow_state(requested_state)
+      else
+        assign_workflow_state(requested_state)
+      end
+    end
+
+    def update_workflow_state(workflow_state)
+      @transient_registration.update_attributes(workflow_state: workflow_state)
+    end
+
+    def assign_workflow_state(workflow_state)
+      @transient_registration.workflow_state = workflow_state
     end
 
     def state_can_navigate_flexibly?(state)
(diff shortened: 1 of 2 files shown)


app/controllers/waste_exemptions_engine/forms_controller.rb
```diff
@@ -79,6 +79,8 @@ def set_workflow_state
       return unless state_can_navigate_flexibly?(@transient_registration.workflow_state)
       return unless state_can_navigate_flexibly?(requested_state)
```
cintamani:
```suggestion
```

app/forms/waste_exemptions_engine/edit_form.rb
```diff
@@ -24,7 +24,10 @@ class EditForm < BaseForm
     # This form has a lot of attributes, so we have to disable the length cop.
     # rubocop:disable Metrics/MethodLength
     def initialize(registration)
+      registration.save!
```
cintamani:
```suggestion
      registration.save! unless registration.persisted?
```

app/forms/waste_exemptions_engine/edit_form.rb
```diff
@@ -24,7 +24,10 @@ class EditForm < BaseForm
     # This form has a lot of attributes, so we have to disable the length cop.
     # rubocop:disable Metrics/MethodLength
```
cintamani:
```suggestion
    # rubocop:disable Metrics/MethodLength Metrics/AbcSize
```

app/controllers/waste_exemptions_engine/forms_controller.rb
```diff
@@ -78,6 +78,7 @@ def setup_checks_pass?
     def set_workflow_state
       return unless state_can_navigate_flexibly?(@transient_registration.workflow_state)
       return unless state_can_navigate_flexibly?(requested_state)
+      return unless @transient_registration.persisted?
 
       @transient_registration.update_attributes(workflow_state: requested_state)
```
cintamani:
```suggestion
      @transient_registration.workflow_state = requested_state
```

app/controllers/waste_exemptions_engine/forms_controller.rb
```diff
@@ -78,6 +78,7 @@ def setup_checks_pass?
     def set_workflow_state
       return unless state_can_navigate_flexibly?(@transient_registration.workflow_state)
       return unless state_can_navigate_flexibly?(requested_state)
+      return unless @transient_registration.persisted?
```
cintamani:
```suggestion
```

app/forms/waste_exemptions_engine/edit_form.rb
```diff
@@ -23,8 +23,12 @@ class EditForm < BaseForm
 
     # This form has a lot of attributes, so we have to disable the length cop.
     # rubocop:disable Metrics/MethodLength
+    # rubocop:disable Metrics/AbcSize
     def initialize(registration)
+      registration.save! unless registration.persisted?
```
Cruikshanks: In todays (yesterdays 😪 ) meeting we said we were not going to touch the edit functionality. But can I assume this is still needed as it fixes something?

cintamani: it does fix something. We said we don't touch it but we need it persisted as it generates too many other fixes. At least this is what I recall 🤔

PR #191, opened 2019-05-23
[WIP] Require wicked pdf in the engine for applications to use

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-223

Given we want to use wicked_pdf superpowers in the WEX back office, in order to require the gem correctly along with its Rails helpers, it must be required in the engine file.


cintamani:
Closing this. Will open a new one.

PR #199, opened 2019-05-29
Tidy up views

cintamani:
Main goal is to improve readability of our views

- Fix identation across all views
- Refactor `if`s on the `error` form class addition to use ERB code inside the class value instead


app/views/waste_exemptions_engine/main_people_forms/_form.html.erb
```diff
@@ -0,0 +1,28 @@
+<%= form_for(@main_people_form) do |f| %>
```
Cruikshanks: It's being picky, but I think this refactoring of the `app/views/waste_exemptions_engine/main_people_forms/new.html.erb` to use a partial is worthy of a PR by itself.

Then this one truly just becomes about whitespace and error class changes.

app/views/waste_exemptions_engine/confirmation_mailer/send_confirmation_email.html.erb
```diff
@@ -15,113 +15,113 @@
...
-              <p style="font-weight: 400; font-size: 16px; line-height: 1; margin: 10px 0 10px 0;">&nbsp;</p>
-            </td>
-            <td width="25%">&nbsp;</td>
-          </tr>
-          <tr>
-            <td bgcolor="#00823B" style="font-family: Helvetica, Arial, sans-serif; margin: 0 0 30px 0;">
-              <p style="font-size: 36px; font-weight: bold; text-align: center; color: #ffffff; line-height: 1.315789474; margin: 30px 0px 15px 0px;">
-                <%= t(".heading.heading") %>
-              </p>
+  <body style="font-family: Helvetica, Arial, sans-serif; font-size: 16px; margin: 0; color: #0b0c0c">
```
Cruikshanks: More a question than feedback. Isn't the `<body>` tag at the same level as the `<head>` tag, and if so shouldn't

- the indentation be applied to both or
- the indentation should match the `<head>` tag

app/views/waste_exemptions_engine/shared/_manual_address.html.erb
```diff
@@ -46,14 +34,10 @@
   </fieldset>
 </div>
 
-<% if form.errors[:city].any? %>
-<div class="form-group form-group-error">
-<% else %>
-<div class="form-group">
-<% end %>
+<div class="form-group <%= "form-group-error" if form.errors[:ciry].any? %>">
```
Cruikshanks: Ooh, typo

```suggestion
<div class="form-group <%= "form-group-error" if form.errors[:city].any? %>">
```

cintamani: Wow!! Great spot!! Wondering why the test suite didn't bring this up 🤔 will check the test coverage next

PR #203, opened 2019-06-03
Allow a service main method to receive no params

cintamani:
Since not all services always needs params, let's make sure we can run services without params. 

Bug came out while I was working on: https://github.com/DEFRA/waste-exemptions-back-office-ta/pull/191/files

### DIFF
--- a/app/services/waste_exemptions_engine/base_service.rb
+++ b/app/services/waste_exemptions_engine/base_service.rb
@@ -2,8 +2,12 @@
 
 module WasteExemptionsEngine
   class BaseService
-    def self.run(attrs)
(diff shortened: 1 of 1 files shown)


app/services/waste_exemptions_engine/base_service.rb
```diff
@@ -2,8 +2,12 @@
 
 module WasteExemptionsEngine
   class BaseService
-    def self.run(attrs)
-      new.run(attrs)
+    def self.run(attrs=nil)
```
cintamani:
```suggestion
    def self.run(attrs = nil)
```

PR #204, opened 2019-06-03
Add area attribute to addresses

Cruikshanks:

The existing service has the ability to lookup the EA admin area for a site using the easting and northing values. This is currently out of action whilst we await a fix (the external service we call changed supplier and we have been awaiting support for what the new query should be).

However it has been spotted in our work on migrating the old data to the new schema that we don't have a field for this. So on the basis we want to retain this data, and that we expect to implement the lookup feature in the new service once the issue is resolved, this change adds a migration to add the required field.

### DIFF
--- a/db/migrate/20190603152123_add_area_to_addresses.rb
+++ b/db/migrate/20190603152123_add_area_to_addresses.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddAreaToAddresses < ActiveRecord::Migration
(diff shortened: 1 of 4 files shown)


irisfaraway:
@Cruikshanks Will we need to display this field in the back office somewhere as well?

Cruikshanks:
Yep, ideally @irisfaraway . I'm gonna handle that with a PR in the Back office project though.

PR #212, opened 2019-06-18
Add HTML w3c test coverage

cintamani:
Part of https://eaflood.atlassian.net/browse/RUBY-388

This fixes some HTML validation issue from W3C HTML5 standards.
It also adds test coverage for valid HTML standard using W3C and VCR. The requests are matched based on the body of the request, which means the cassettes will regenerate and validate the content every time the view code changes.


spec/requests/waste_exemptions_engine/contact_address_lookup_forms_spec.rb
```diff
@@ -3,7 +3,7 @@
 require "rails_helper"
 
 module WasteExemptionsEngine
-  RSpec.describe "Contact Address Lookup Forms", type: :request, vcr: true do
```
cintamani: This tag here was never useful since VCR is not configured for tags to work yet, hence I have removed it

spec/support/shared_examples/requests/get_form.rb
```diff
@@ -18,6 +18,12 @@
         get good_request_path
         expect(response.code).to eq("200")
       end
+
+      it "returns W3C valid HTML content" do
+        get good_request_path
+
+        expect(response.body).to have_valid_html
+      end
```
cintamani: This is the change that made me like the choice of the shared examples :D

spec/requests/waste_exemptions_engine/contact_address_lookup_forms_spec.rb
irisfaraway: Is it not? I have been using `bundle exec -t vcr` when regenerating the cassettes and it seems to work.

cintamani: Well, when the tag is used then we don't need to deal with the `use_cassettes` methods while we do instead. Let me rollback this change then and let's tackle it separately :)

spec/requests/waste_exemptions_engine/contact_address_lookup_forms_spec.rb
```diff
@@ -3,7 +3,7 @@
...
+  RSpec.describe "Contact Address Lookup Forms", type: :request do
```
cintamani:
```suggestion
  RSpec.describe "Contact Address Lookup Forms", type: :request, vcr: true do
```

spec/requests/waste_exemptions_engine/contact_address_lookup_forms_spec.rb
irisfaraway: Cool :) Can we apply the `vcr` tag to the new w3c validation test as well?

cintamani:
@irisfaraway noticed that the Cassettes are regenerated all the times. This is because the URI actually changes all the times (for the token). Putting this back in WIP to sort out that issue.

spec/support/html_matchers.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require "rspec/expectations"
+require "w3c_validators"
+
+RSpec::Matchers.define :have_valid_html do
+  match do |actual|
+    @validator = W3CValidators::NuValidator.new
+    vcr_options = { match_requests_on: %i[html_body_content], record: :once}
```
cintamani:
```suggestion
    vcr_options = { match_requests_on: %i[html_body_content], record: :once }
```

cintamani:
@irisfaraway I have changed the VCR configuration so one cassette is generated per-request. This will avoid the test suite being slow because of the body request matching having to scan all the registrations in the file. 
I also had to fix the clean method in the custom matcher as the content wasn't cleaned up properly. 
Ready to review again :)

spec/support/shared_examples/requests/get_form.rb
```diff
@@ -18,6 +18,12 @@
         get good_request_path
         expect(response.code).to eq("200")
       end
+
+      it "returns W3C valid HTML content" do
```
cintamani:
```suggestion
      it "returns W3C valid HTML content", vcr: true do
```

spec/support/vcr.rb
```diff
@@ -21,9 +21,12 @@
 
   c.register_request_matcher :html_body_content do |request_one, request_two|
     def clean(string)
-      string.gsub(/\"[^\"]+\"/, "")
+      string.gsub(/""/, "\"FOO\"").gsub(/"[^"]*"/, "\"Classes\"").gsub(/></, ">BAR<").gsub(/>[^<>]*</, ">Info<")
```
irisfaraway: I think it'd be useful to have a comment here explaining what this line is doing.

spec/requests/waste_exemptions_engine/exemptions_forms_spec.rb
```diff
@@ -4,7 +4,10 @@
 
 module WasteExemptionsEngine
   RSpec.describe "Exemptions Forms", type: :request do
-    before(:context) { create_list(:exemption, 5) }
+    before(:context) do
+      WasteExemptionsEngine::Exemption.delete_all
```
irisfaraway: What does this change do? Is this so we can be sure we are getting a certain number of exemptions?

spec/support/html_matchers.rb
```diff
@@ -6,8 +6,11 @@
 RSpec::Matchers.define :have_valid_html do
   match do |actual|
     @validator = W3CValidators::NuValidator.new
-    vcr_options = { match_requests_on: %i[uri method html_body_content], record: :new_episodes }
-    @results = VCR.use_cassette("w3c_valid_content", vcr_options) do
+    vcr_options = { match_requests_on: %i[html_body_content], record: :once}
+
+    cassette_name = method_missing(:class).metadata[:rerun_file_path].split("/").last.gsub(".rb", "")
```
irisfaraway: Might be useful to have a comment with the expected result as there's quite a lot happening on this line.

spec/requests/waste_exemptions_engine/exemptions_forms_spec.rb
cintamani: Yes. Especially for Exemptions, the HTML is generated based on the content of the database. I have noticed that those tests populating the DB with a before context were flakey, and the reason was in the number of exemptions that wasn't always 5. 
I had to put a dot somewhere so I haven't spent too long on finding out which test is leaving the DB dirty and why the DB cleaner is not picking it up. One for the next time :)

cintamani:
In regard to the VCR failure, that's per-design. 
In order to `fix` it, the cassette will need to be destroyed and regenerated by running the test again. 
Unfortunately, there is no way right now for me to tell VCR to return a custom error message when a custom matcher is used, that's still an [open issue](https://github.com/vcr/vcr/issues/610) on the repo.

irisfaraway:
@cintamani Comments look good - much clearer now!

Is there any way to just regenerate the cassette automatically when the HTML has changed and see if the updated version validates, rather than make whoever is running the test delete and recreate it? I don't think the test needs to fail just if the structure of the page changes, only if the structure itself is now invalid.

cintamani:
We had a phone discussion about the issue, and we agreed that: 

The problem is that there is no easy way to distinguish between a "I cannot find a registered call for this" and "Your HTML changed". 
Right now we can't find a solution we would be happy with that does not require the deletion of the cassette for it to be re-registered. 
We could write some code that infiltrates into VCR and checks if the cassette as a registered call for the specific HTML before running the test and custom-fail with the error we want to see, but that would be a tricky change and we feel like is not worth spending the time doing that right now. 
An issue has been created [here](https://github.com/DEFRA/ruby-services-team/issues/42) instead to track the problem. 
The aim is to go out of the door with this change and see how much pain the maintenance of these cassettes is giving us and come out with a solution later on.

PR #214, opened 2019-06-19
Auto-determine grid reference for site address

Cruikshanks:

Whilst working on the migration of data from the old schema to the new schema we identified (remembered!) that the existing service automatically determines the grid reference if an address is used for the site.

This information is then what gets exported to the Electronic Public Register, and is used in its 'site is within x miles of postcode' search.

We will currently attempt to determine the x & y for manually entered site addresses in order to support the EA admin area lookup (to be implemented), but we now know we also need to be determining the grid reference.

This change updates the `create_from_x()` methods in the `transient_address` model to also include an attempt to determine the grid reference for a site.


spec/models/waste_exemptions_engine/transient_address_spec.rb
```diff
@@ -3,7 +3,7 @@
 require "rails_helper"
 
 module WasteExemptionsEngine
-  RSpec.describe TransientAddress, type: :model do
+  RSpec.describe TransientAddress, type: :model, vcr: true do
```
irisfaraway: Can we move `vcr: true` to just the relevant context?

Cruikshanks: NP. That was me quickly looking at another spec and assuming it needed to be at the top!

PR #220, opened 2019-06-26
Store a single version of registration and its relations

irisfaraway:
When a registration's address changes, or some other object belonging to the registration changes,  we want to store that the registration was updated.

To accomplish this, we've decided to store tracking addresses alone and instead force a new version to be saved for the registration when an edit is completed. We'll also store the "full" registration data with all the related object data as JSON so we can get a full record.


app/models/waste_exemptions_engine/registration.rb
```diff
@@ -18,5 +18,13 @@ class Registration < ActiveRecord::Base
       through: :registration_exemptions,
       source: :exemption
     )
+
+    private
+
+    def json_for_version
+      to_json(include: %i[addresses
```
cintamani: I love this method <3

Cruikshanks:
> I don't see anything wrong nor blockers for this.
> However, can I suggest that we create a separate object that deals with the entire knowledge of how to compare a registration with a transient object and decide if it has changed or not? It feels to me that we are putting too much knowledge in one object with this change 🤔

That's actually a blummin' brilliant suggestion. However I'm conscious of current time frames so I'm still happy to approve, but one for the issues list to refactor if you don't think you can squeeze it in @irisfaraway

irisfaraway:
@cintamani @Cruikshanks I think it's a really good suggestion and agree we should do it! However I'm tempted to pop it into the issues list for now and plan to pick it up soon, purely because Shipping Is Coming and I'd like to get this into QA and signed off ASAP.

PR #222, opened 2019-07-02
Add feature toggles class to reuse in apps

cintamani:
This adds a FeatureToggles model to the engine that build the structure around dealing with feature toggles.

### DIFF
--- a/app/models/waste_exemptions_engine/feature_toggle.rb
+++ b/app/models/waste_exemptions_engine/feature_toggle.rb
@@ -5,7 +5,7 @@ require "yaml"
 module WasteExemptionsEngine
   class FeatureToggle
     def self.active?(feature_name)
-      feature_toggles[feature_name] && feature_toggles[feature_name][:active]
+      feature_toggles[feature_name] && feature_toggles[feature_name][:active] == true
     end
 
     class << self
@@ -22,7 +22,7 @@ module WasteExemptionsEngine
       end
 
       def file_path
-        Rails.root.join("config/feature_toggles.yaml")
+        Rails.root.join("config/feature_toggles.yml")
       end
     end
   end
--- a/app/models/waste_exemptions_engine/feature_toggle.rb
+++ b/app/models/waste_exemptions_engine/feature_toggle.rb
@@ -8,18 +8,22 @@ module WasteExemptionsEngine
       feature_toggles[feature_name] && feature_toggles[feature_name][:active]
     end
 
-    private
+    class << self
+      private
 
-    def self.feature_toggles
-      @@feature_toggles ||= load_feature_toggles
-    end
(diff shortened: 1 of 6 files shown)


spec/dummy/config/database.yml
```diff
@@ -6,8 +6,8 @@ default: &default
   pool: 5
   host: <%= ENV["PG_HOST"] || "localhost" %>
   port: <%= ENV["PG_PORT"] || 5432 %>
-  user: <%= ENV["PG_USERNAME"] || "vagrant" %>
-  password: <%= ENV["PG_PASSWORD"] || "vagrant" %>
+  user: cintamanipuddu
```
cintamani:
```suggestion
  user: <%= ENV["PG_USERNAME"] || "vagrant" %>
```

spec/dummy/config/database.yml
```diff
@@ -6,8 +6,8 @@ default: &default
...
+  password:
```
cintamani:
```suggestion
  password: <%= ENV["PG_PASSWORD"] || "vagrant" %>
```

app/models/waste_exemptions_engine/feature_toggle.rb
```diff
@@ -0,0 +1,29 @@
...
+        @@feature_toggles ||= load_feature_toggles
+      end
+      # rubocop:enable Style/ClassVars
+
+      def load_feature_toggles
+        HashWithIndifferentAccess.new(YAML.load_file(file_path))
+      end
+
+      def file_path
+        Rails.root.join("config/feature_toggles.yaml")
```
irisfaraway: This may be the nitpickiest thing ever but we usually use a .yml suffix instead of .yaml. So maybe worth a switch for consistency's sake.

Cruikshanks: ooh, good spot! 🔍 🕵️‍♀️

cintamani: Ah! Nice one! Will change it!

cintamani:
@irisfaraway good point! Will test that and see! I would expect it to return `false` thought, and not to fail

spec/models/waste_exemptions_engine/feature_toggle_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+          end
+        end
+
+        context "when it is configured as not active" do
+          it "returns false" do
+            expect(described_class.active?("not_active_test_feature")).to be_falsey
+          end
+        end
+
+        context "when the feature toggle contains a typo in the rreturn value" do
```
Cruikshanks: Just a typo

```suggestion
        context "when the feature toggle contains a typo in the return value" do
```

PR #225, opened 2019-07-05
Show error messages in line with address select fields

cintamani:
While working on setting up spec tests for the errors we show to the user, we found that the address lookup pages were the only pages not showing the message in line with the field.
Seems like some of the code necessary for the field to show was there but pointing to `addresses` rather than `temp_addresses`. I decided to just adapt the code to work in both cases when there are errors on `addresses` and when there are in `temp_addresses`.

Test coverage will come in a separate PR. 

Screenshot 2019-07-05 at 12 01 06

### DIFF
--- a/app/views/waste_exemptions_engine/shared/_select_address.html.erb
+++ b/app/views/waste_exemptions_engine/shared/_select_address.html.erb
@@ -1,5 +1,13 @@
-<div class="form-group <%= "form-group-error" if form.errors[:addresses].any? %>">
+<div class="form-group <%= "form-group-error" if form.errors[:addresses].any? || form.errors[:temp_address].any? %>">
   <fieldset id="temp_address">
+    <% if form.errors[:temp_address].any? %>
+      <span class="error-message"><%= form.errors[:temp_address].join(", ") %></span>
+    <% end %>
+
+    <% if form.errors[:addresses].any? %>
+      <span class="error-message"><%= form.errors[:addresses].join(", ") %></span>
+    <% end %>
+
     <%= f.label :temp_address, t(".address_label"), class: "form-label" %>
     <%= f.select(:temp_address,
                  options_for_select(form.temp_addresses.map { |a| [a["address"], a["uprn"]] }, form.temp_address),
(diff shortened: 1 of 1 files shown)


app/views/waste_exemptions_engine/shared/_select_address.html.erb
```diff
@@ -1,5 +1,13 @@
-<div class="form-group <%= "form-group-error" if form.errors[:addresses].any? %>">
+<div class="form-group <%= "form-group-error" if form.errors[:addresses].any? || form.errors[:temp_address].any? %>">
   <fieldset id="temp_address">
+    <% if form.errors[:temp_address].any? %>
+      <span class="error-message"><%= form.errors[:temp_address].join(", ") %></span>
```
irisfaraway: Will there ever be a scenario where there are errors in both `temp_address` and `addresses`? If so, should we be joining them rather than displaying them in separate `span` tags?

Cruikshanks:
Just need to remember we are maintaining multiple services. So if we have the issue here, we probably also have the issue in [waste-carriers-engine](https://github.com/DEFRA/waste-carriers-engine).

Would you mind double checking and making the same changes there (if required)?

app/views/waste_exemptions_engine/shared/_select_address.html.erb
cintamani: I actually have no clue about where the `addresses` is coming from 😓so I have left it working for both. Testing it in the normal flow showed that just "temp_address" is a field. Probably some old code was doing that? 🤔 @Cruikshanks can you recall where the `addresses` field is coming from in this context?

Cruikshanks: Because a long, long time ago in a galaxy far, far away the attribute name was `addresses` https://github.com/DEFRA/waste-exemptions-engine/blob/ed44ad18fd23007b02f184a06e8bcd6f26b36903/config/locales/forms/operator_address_forms/en.yml#L23

So feel free to kill it! 😁

cintamani: Alan as confirmed that this is legacy, so removing references to the `addresses` field :)

PR #226, opened 2019-07-05
Test coverage for error messages

cintamani:
Following a refactoring of our view, we have noticed that there is no test coverage on the shown of an error message to the users. This PR adds common validation messages coverage for our request spec.

### DIFF
--- a/spec/support/shared_examples/requests/post_form.rb
+++ b/spec/support/shared_examples/requests/post_form.rb
@@ -6,6 +6,7 @@ RSpec.shared_examples "POST form" do |form_factory, path, empty_form_is_valid = 
   let(:incorrect_form) { build(incorrect_workflow_state) }
   let(:post_request_path) { "/waste_exemptions_engine#{path}" }
   let(:form_data) { { override_me: "Set :form_data in the calling spec." } }
+  let(:invalid_form_data) { { override_me: "Set :invalid_form_data in the calling spec." } }
 
   describe "POST #{form_factory}" do
     context "when the form is not valid", unless: empty_form_is_valid do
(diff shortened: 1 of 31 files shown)


spec/support/shared_examples/requests/post_form.rb
```diff
@@ -20,6 +21,21 @@
...
+      it "includes validation errors for the form data" do
+        invalid_form_data.each do |invalid_data|
+          post post_request_path, form_factory => invalid_data.merge(token: correct_form.token)
+
+          invalid_form = build(form_factory, **invalid_data)
+          invalid_form.validate
+
+          invalid_form.errors.messages.values.flatten.each do |error_message|
+            # We include error messages twice, but rspec have no built-in an include twice method yet.
+            # Hence, the scan will make sure we matche the message twice in the rendered page.
```
irisfaraway:
```suggestion
            # Hence, the scan will make sure we match the message twice in the rendered page.
```

Also, in this case, why not `= 2`?

spec/support/shared_examples/requests/post_form.rb
```diff
@@ -20,6 +21,21 @@
...
+
+      it "includes validation errors for the form data" do
+        invalid_form_data.each do |invalid_data|
+          post post_request_path, form_factory => invalid_data.merge(token: correct_form.token)
+
+          invalid_form = build(form_factory, **invalid_data)
+          invalid_form.validate
+
+          invalid_form.errors.messages.values.flatten.each do |error_message|
+            # We include error messages twice, but rspec have no built-in an include twice method yet.
```
irisfaraway:
```suggestion
            # We include error messages twice, but RSpec has no built-in "include twice" method yet.
```

Spotted a typo and then thought this might be a bit clearer.

Cruikshanks:
There's no point me approving because when you correct the typo's it'll dismiss it. But when that's done I'll be all over this with ♥️, 💋, and 🤗 !

spec/support/shared_examples/requests/post_form.rb
cintamani: It is because in some views the same error message text appears as page text for explaining to the user what to do. Hence, in order to keep the maintainability of this test low, I have decided that testing that we show "at least 2" of those error messages is good enough :)

PR #231, opened 2019-07-09
Externalise logic for comparing Registration rerlated objects

cintamani:
From https://github.com/DEFRA/ruby-services-team/issues/44

Create a service to compare registration and edit registration related objects and decide if the registration has been edited or not.


app/services/waste_exemptions_engine/edit_completion_service.rb
```diff
@@ -19,8 +19,6 @@ def find_original_registration
     end
 
     def copy_data_from_edit_registration
-      related_objects_changed?
```
cintamani: I am totally unaware of why this code was here, and hence I am going to wait for Iris feedback before merging this code

cintamani: Ok so it took me a while, and failing tests helped, but I figured. 
The issue is that when copying related objects in the `@registration` we are effectively saving the registration. Hence when then comparing the object at the end no changes arise. 
In order to fix this, I have changed the logic a bit, by checking first of all if the object has changed at all, and if it did then I do the copy over. In the end, I always create a new version.

irisfaraway:
@cintamani Do you still want me to review this? Looks like the build is broken at the moment

cintamani:
Yep! Missed that! no rush on this. Will fix it next week :)

cintamani:
Closing this for now.

PR #232, opened 2019-07-10
Update date in privacy policy PDF

Cruikshanks:
Pull request #230 changed the last updated date in the privacy policy HTML view. But we completely forgot that this is used to generate a static copy of the policy as a PDF file.

Because the content is static, we made the decision to generate the static file and attach that to each email, rather than generate it each time.

So this change updates the date in the overlooked PDF file, so now what we attach to the email is correct.


Cruikshanks:
I know! when I left last night and @andrewhick told me it wasn't working I was bamboozled.

A nights sleep and 2 minutes later 🤦‍♂🤦‍♂🤦‍♂🤦‍♂🤦‍♂🤦‍♂🤦‍♂🤦‍♂!

PR #237, opened 2019-07-24
Use DB Pessimistic locking to prevent race conditions

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-523

In order to avoid race conditions creating broken registration records, this PR adds a test coverage which take advantage of Ruby Threads to recreate the race condition.
To fix the race condition, we have decided to use the `transient_registration` workflow_state to run an update on the object which will make it possible to take advantage of the https://api.rubyonrails.org/classes/ActiveRecord/Locking/Pessimistic.html.

### DIFF
--- a/spec/services/waste_exemptions_engine/registration_completion_service_spec.rb
+++ b/spec/services/waste_exemptions_engine/registration_completion_service_spec.rb
@@ -11,18 +11,23 @@ module WasteExemptionsEngine
 
     describe "#complete" do
       it "is idempotent" do
-        expect(WasteExemptionsEngine::Registration.count).to eq(0)
+        # FIXME: Some test is leaving the db dirty. Hence we need a count and expect 1 extra.
+        # https://github.com/DEFRA/ruby-services-team/issues/54
+        initial_count = WasteExemptionsEngine::Registration.count
+
         RegistrationCompletionService.new(new_registration).complete
 
         expect { RegistrationCompletionService.new(new_registration).complete }.to raise_error(StandardError)
 
-        expect(WasteExemptionsEngine::Registration.count).to eq(1)
+        expect(WasteExemptionsEngine::Registration.count).to eq(initial_count + 1)
       end
 
       context "when a race condition calls the service twice" do
         # rubocop:disable Lint/HandleExceptions
         it "generates only one record and fail to execute subsequent calls" do
-          expect(WasteExemptionsEngine::Registration.count).to eq(0)
+          # FIXME: Some test is leaving the db dirty. Hence we need a count and expect 1 extra.
+          # https://github.com/DEFRA/ruby-services-team/issues/54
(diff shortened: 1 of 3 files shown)


spec/services/waste_exemptions_engine/registration_completion_service_spec.rb
```diff
@@ -10,6 +10,55 @@ module WasteExemptionsEngine
...
+          ensure
+            ActiveRecord::Base.connection_pool.disconnect!
+          end
+
+          should_wait = false
+          threads.each(&:join)
+
+          expect(WasteExemptionsEngine::Registration.count).to eq(1)
+
+          # Clean up after the threads have executed
```
irisfaraway: DatabaseCleaner should be configured to clean up around each example, so I'm not sure we need to do this ourselves - was this because of a specific issue?

spec/services/waste_exemptions_engine/registration_completion_service_spec.rb
```diff
@@ -10,6 +10,55 @@ module WasteExemptionsEngine
     let(:registration_completion_service) { RegistrationCompletionService.new(new_registration) }
 
     describe "#complete" do
+      it "is idempotent" do
+        expect(WasteExemptionsEngine::Registration.count).to eq(0)
+        RegistrationCompletionService.new(new_registration).complete
+
+        expect { RegistrationCompletionService.new(new_registration).complete }.to raise_error
```
irisfaraway: RSpec does raise this message when we run this test:

```
WARNING: Using the `raise_error` matcher without providing a specific error or message risks false positives, since `raise_error` will match when Ruby raises a `NoMethodError`, `NameError` or `ArgumentError`, potentially allowing the expectation to pass without even executing the method you are intending to call. Actual error raised was #<RuntimeError: Can't modify frozen hash>. Instead consider providing a specific error class or message.
```

I agree with it, as it looks like it'd be pretty quick to modify the test to include this specific error.

spec/services/waste_exemptions_engine/registration_completion_service_spec.rb
cintamani: This is because we use a transaction strategy, and Database cleaner can't do transactions with Threads, as they happen outside of the rspec process

spec/services/waste_exemptions_engine/registration_completion_service_spec.rb
cintamani: Nice one! Will fix that

spec/services/waste_exemptions_engine/registration_completion_service_spec.rb
irisfaraway: @cintamani Got it 👍 I knew there must have been a reason!

PR #239, opened 2019-07-24
Change position of the Exemptions form

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-457

Change the state machine and fix tests so that the exemptions screen will now appear just after the location screens.

Other fixes were necessary on the test suite for all this to run. First thing that went wrong with them, is that removing the exemptions form between the site address and check your answers screen was generating an issue with the helper we use to fetch the correct `previous state` of a form. The code that was there was doing very little on fetching the correct previous status, was not accounting for multiple options, like is the case when dealing with the site_address, and was using "next" AASM events rather than "back" ones, so it was actually working just by chances rather than correct logic.
Once that was fixed to return the correct previous state, it was now breaking the logic for "flexibly navigatable" tests in the "get form" share examples. This is becase we rely on the get_previous_state to return a result for forms that should not have any previous state, like is the case for not flexibly navigatable forms, hence the fixes there.

### DIFF
--- a/spec/support/helpers/workflow_states.rb
+++ b/spec/support/helpers/workflow_states.rb
@@ -7,9 +7,11 @@ module Helpers
     end
 
     def self.can_navigate_flexibly_to_state?(state)
+      return false unless state_can_navigate_flexibly?(state)
+
       previous_state = previous_state(WasteExemptionsEngine::NewRegistration.new(workflow_state: state))
 
-      previous_state && state_can_navigate_flexibly?(previous_state) && state_can_navigate_flexibly?(state)
+      previous_state && state_can_navigate_flexibly?(previous_state)
     end
 
     def self.state_can_navigate_flexibly?(state)
@@ -17,21 +19,30 @@ module Helpers
       form_class.can_navigate_flexibly?
     end
 
+    # The purpose of this method is to dynamically find a valid previous state for a given transient registration's state.
+    # In order to do that, we get the data of the state machine of the transient registration class, and we search in all
+    # back event which handles moving from a transient registration's current state to a previous valid state.
     def self.previous_state(transient_registration)
       state_machine = transient_registration.class.aasm
 
       state_machine.events.each do |event|
+        # We are only interested in :back events. Those are the only that define a valid previous state for a
+        # transient registration
         next unless event.name == :back
 
         previous_state_transition = event.transitions.find do |transition|
+          # Once we find a valid transition, we have to make sure such transition can be run in the current object.
+          # Some transitions have extra `:if:` and `:unless:` rules that define on which object they are valid to run,
+          # the `#allowed?` method will run those checks on the give object for the given transaction
           transition.allowed?(transient_registration) && transition.from == transient_registration.aasm.current_state
         end
 
+        # We exit the method and return the first valid result
         return previous_state_transition.to if previous_state_transition
       end
 
-      return if transient_registration.registration_complete_form?
-
+      # If no result is found, then there is something logically wrong. One reason might be that the current state is not a
+      # flexibly navigatable one.
       raise(StandardError, "No previous state found for #{transient_registration.aasm.current_state}")
     end
   end
--- a/spec/support/helpers/workflow_states.rb
+++ b/spec/support/helpers/workflow_states.rb
@@ -8,7 +8,8 @@ module Helpers
 
     def self.can_navigate_flexibly_to_state?(state)
       previous_state = previous_state(WasteExemptionsEngine::NewRegistration.new(workflow_state: state))
-      state_can_navigate_flexibly?(previous_state) && state_can_navigate_flexibly?(state)
+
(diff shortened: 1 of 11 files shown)


spec/support/helpers/workflow_states.rb
```diff
@@ -20,14 +21,20 @@ def self.previous_state(transient_registration)
...
-          transition.to == transient_registration.aasm.current_state
+          transition.allowed?(transient_registration) && transition.from == transient_registration.aasm.current_state
         end
 
-        return previous_state_transition.from if previous_state_transition
+        return previous_state_transition.to if previous_state_transition
       end
 
-      raise StandardError, "No previous state found for #{transient_registration.aasm.current_state}"
+      :start_form
```
cintamani:
```suggestion
```

spec/support/helpers/workflow_states.rb
```diff
@@ -20,14 +21,19 @@ def self.previous_state(transient_registration)
...
-          transition.to == transient_registration.aasm.current_state
+          transition.allowed?(transient_registration) && transition.from == transient_registration.aasm.current_state
         end
 
-        return previous_state_transition.from if previous_state_transition
+        return previous_state_transition.to if previous_state_transition
       end
 
-      raise StandardError, "No previous state found for #{transient_registration.aasm.current_state}"
+
```
cintamani:
```suggestion
```

spec/support/helpers/workflow_states.rb
```diff
@@ -20,14 +21,18 @@ def self.previous_state(transient_registration)
       state_machine = transient_registration.class.aasm
 
       state_machine.events.each do |event|
```
irisfaraway: I'd love to get some comments on this method! Otherwise it does require some digging into the AASM API to understand what is going on.

spec/support/helpers/workflow_states.rb
```diff
@@ -7,31 +7,42 @@ def self.permitted_states(transient_registration)
...
         end
 
+        # We exit the method and return the first valid result
         return previous_state_transition.to if previous_state_transition
       end
 
-      return if transient_registration.registration_complete_form?
-
+      # If no result is found, then there is something logically wrong. One reason might be that the current state is not a
+      # flexibly navigatable one.
```
irisfaraway: This isn't the case – whether something is "flexibly navigatable" is just about whether it can be skipped to out-of-sequence, rather than whether it has any previous pages.

For example, the declaration form is not flexible as we want to force users to go through the correct forms before accessing it. But it does have a previous page, "check your answers".

I think "One reason might be that the current state is the start of the user journey, or an inescapable end point" would be more accurate.

cintamani: I am so confused :/

cintamani: Will apply that words change! make sense

spec/support/helpers/workflow_states.rb
```diff
@@ -7,27 +7,44 @@ def self.permitted_states(transient_registration)
...
+          transition.allowed?(transient_registration) && transition.from == transient_registration.aasm.current_state
         end
 
-        return previous_state_transition.from if previous_state_transition
+        # We exit the method and return the first valid result
+        return previous_state_transition.to if previous_state_transition
       end
 
-      raise StandardError, "No previous state found for #{transient_registration.aasm.current_state}"
+      # If no result is found, then there is something logically wrong. One reason might be that the current state is
```
cintamani:
```suggestion
      # If no result is found, then there is something logically wrong. One reason might be that the current state is
```

spec/support/helpers/workflow_states.rb
```diff
@@ -7,27 +7,44 @@ def self.permitted_states(transient_registration)
...
+      # not a flexibly navigatable one.
```
cintamani:
```suggestion
      #  the start of the user journey, or an inescapable end point.
```

spec/support/helpers/workflow_states.rb
```diff
@@ -7,27 +7,44 @@ def self.permitted_states(transient_registration)
...
         end
 
-        return previous_state_transition.from if previous_state_transition
+        # We exit the method and return the first valid result
+        return previous_state_transition.to if previous_state_transition
       end
 
-      raise StandardError, "No previous state found for #{transient_registration.aasm.current_state}"
+      # If no result is found, then there is something logically wrong. One reason might be that the current state is
+      #  the start of the user journey, or an inescapable end point.
```
cintamani:
```suggestion
      # the start of the user journey, or an inescapable end point.
```

spec/support/helpers/workflow_states.rb
irisfaraway: It is confusing! I created an issue which hopefully has a bit more explanation: https://github.com/DEFRA/ruby-services-team/issues/56

irisfaraway:
Argh, unapprove, Travis is failing.

cintamani:
Oh shut :joy: shall we pair?

PR #242, opened 2019-07-26
Add existing forms to renewals workflow

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-516

We have created a new state machine for the renewals journey in https://github.com/DEFRA/waste-exemptions-engine/pull/241. But the full renewals journey needs to include lots of forms which are already used in the new registration journey.

This PR covers adding those existing forms to the renewals workflow.


irisfaraway:
This branch has nothing on it and has already been rebased a million times so I'm going to start fresh for the sake of a cleaner PR.

PR #245, opened 2019-07-29
Generate unique renewal token and add constraint for magic link

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-514

- Add `renew_token` field to registration tables. Add unique index to prevent race conditions.
- Add service to generate a valid renew token, `GenerateRenewTokenService`.
- Modify routes and Renew controller to behave as magic link route. Added autenticity check service and used to redirect the user to a not authorised page.


app/services/waste_exemptions_engine/generate_renew_token_service.rb
```diff
@@ -0,0 +1,29 @@
...
+      registration.renew_token = generate_token(payload)
+
+      registration.save!
+    end
+
+    private
+
+    def exp
+      # Expires in 30 days
+      Time.now.to_i + 30 * 24 * 3600
```
cintamani:
```suggestion
      Time.now.to_i + 60 * 24 * 3600
```

app/services/waste_exemptions_engine/generate_renew_token_service.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      registration.renew_token = generate_token(payload)
+
+      registration.save!
+    end
+
+    private
+
+    def exp
+      # Expires in 30 days
```
cintamani:
```suggestion
      # Expires in 60 days
```

app/services/waste_exemptions_engine/concerns/can_process_jwt_tokens.rb
```diff
@@ -0,0 +1,29 @@
...
+      def generate_token(payload)
+        JWT.encode(payload, signature, algorithm)
+      end
+
+      def decode_token(token)
+        JWT.decode(token, signature, true, algorithm: algorithm)
+      end
+
+      def signature
+        # TODO: Decide on a more secure algorithm and fix me
```
Cruikshanks: Think we can remove the TODO's now we have our response from security

app/services/waste_exemptions_engine/generate_renew_token_service.rb
```diff
@@ -0,0 +1,29 @@
...
+      registration.renew_token = generate_token(payload)
+
+      registration.save!
+    end
+
+    private
+
+    def exp
+      # Expires in 60 days
+      Time.now.to_i + 60 * 24 * 3600
```
Cruikshanks: We need to get the 60 days from an env var. Though we're unlikely to change it in the life of the app, being able to configure it differently will help us with testing (QA rather than unit testing).

app/services/waste_exemptions_engine/concerns/can_process_jwt_tokens.rb
cintamani:
```suggestion
```

app/services/waste_exemptions_engine/concerns/can_process_jwt_tokens.rb
```diff
@@ -0,0 +1,29 @@
...
+        Rails.application.secrets.secret_key_base
+      end
+
+      def algorithm
+        # TODO: Decide on a more secure algorithm and fix me
```
cintamani:
```suggestion
```

spec/dummy/config/application.rb
```diff
@@ -32,6 +32,9 @@ class Application < Rails::Application
     # Ensure the tests only pass when all the required tralations are present.
     config.action_view.raise_on_missing_translations = true
 
+    # Change authomatic expire of renew's magic link token
+    config.renew_token_expires_in_days = ENV["RENEW_TOKEN_EXPIRES_IN_DAYS"] || 60
```
irisfaraway: Is 60 days right? I don't think this ticket specifies the token expiry, but RUBY-496 says the grace period ends one month after the original registration expires. If we go with days then we could have someone receive the email on 1 July, expire on 1 August and then be unable to renew on 29 August – so they'd be missing out on 2 days.

If we do decide to go with days I think we need to make sure everyone is clear on this so we are telling the user the right information. (eg "this will expire in 60 days" instead of "this will expire in 2 months").

spec/support/helpers/generate_renew_token_helper.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require "jwt"
+
+module GenerateRenewTokenHelper
```
irisfaraway: Our other spec helpers are namespaced (eg `Helpers::GenerateRenewTokenHelper`) so I'd suggest doing that here as well 👍

spec/support/helpers/generate_renew_token_helper.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def generate_renew_token_without_updating_registration(registration)
+    payload = { data: { registration_id: registration.id }, exp: Time.now.to_i - 100 }
+
+    JWT.encode(payload, Rails.application.secrets.secret_key_base, "HS384")
+  end
+end
+
+RSpec.configure do |config|
+  config.include GenerateRenewTokenHelper
```
irisfaraway: Is this necessary? I think the other helpers are included automatically so would expect this to be the same.

spec/dummy/config/application.rb
cintamani: Good point. I'll ask about this in tomorrow's standup :)

spec/support/helpers/generate_renew_token_helper.rb
cintamani: Will change this :)

spec/dummy/config/application.rb
Cruikshanks: Whilst I think about it, it might be prudent to split the config into 2 as well so we have a bit of flexibility and it's clearer in the code what's going on.

So we'd have something like

- `REGISTRATION_RENEWAL_WINDOW=1`
- `REGISTRATION_GRACE_WINDOW=1`

The JWT token expiry would then be the sum of both values

> This idea is copied from Waste Carriers which splits the config

irisfaraway: Ooh, I agree with this ☝️

cintamani: Me too! But can I suggest we use the existing `FIRST_RENEWAL_EMAIL_REMINDER_DAYS`? Or, even better, `x days until expires` + `REGISTRATION_GRACE_WINDOW`?

cintamani: If I understood it correctly, what we want in reality is a token that is "active" from the moment it is generated until the last possible day before the renewal + a grace period, so let's code that :)

irisfaraway: @cintamani Sounds good - `FIRST_RENEWAL_EMAIL_REMINDER_DAYS + REGISTRATION_GRACE_WINDOW` makes sense to me! 👍

spec/dummy/config/application.rb
```diff
@@ -33,8 +33,8 @@ class Application < Rails::Application
     config.action_view.raise_on_missing_translations = true
 
     # Change automatic expire of renew's magic link token
-    config.renew_token_expires_in_days = ENV["RENEW_TOKEN_EXPIRES_IN_DAYS"] || 60
-
+    config.registration_renewal_grace_window = ENV["REGISTRATION_RENEWAL_GRACE_WINDOW"] || 30
+    config.first_renewal_email_reminder_days = ENV["FIRST_RENEWAL_EMAIL_REMINDER_DAYS"]
```
irisfaraway: We'll need to add this to the .env example as well :)

spec/dummy/.env.example
```diff
@@ -39,3 +39,6 @@ COMPANIES_HOUSE_API_KEY=
 
 # Generate PDF's in front-office
 USE_XVFB_FOR_WICKEDPDF=true
+
+# Grace window value in days from when a registration expires
+REGISTRATION_RENEWAL_GRACE_WINDOW=30
```
cintamani:
```suggestion
REGISTRATION_RENEWAL_GRACE_WINDOW=30
FIRST_RENEWAL_EMAIL_REMINDER_DAYS=28
```

PR #246, opened 2019-07-30
Add existing forms to renewals workflow

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-516

If a user is making changes, they will go through a very similar workflow to the new registration journey, with minor differences at the start and end.

This PR adds those existing forms to the state machine for the renewals workflow.


irisfaraway:
From this point on will be a lot of refactoring... brace for rebasing!

irisfaraway:
Attempted to refactor this into some nested concerns but couldn't get it to work. Agreed we'd log an issue to revisit this later and see if we can reduce the duplication.

PR #247, opened 2019-07-31
Complete renewal and create new registration

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-517

When a user reaches the "renewal complete" page, we should create a new registration with the data they just confirmed.

The renewing_registration record should also be deleted from the database.

### DIFF
--- a/app/services/waste_exemptions_engine/registration_completion_service.rb
+++ b/app/services/waste_exemptions_engine/registration_completion_service.rb
@@ -1,12 +1,9 @@
 # frozen_string_literal: true
 
 module WasteExemptionsEngine
-  class RegistrationCompletionService
-    def initialize(transient_registration)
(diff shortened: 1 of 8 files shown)


app/services/waste_exemptions_engine/registration_completion_service.rb
```diff
@@ -1,12 +1,9 @@
 # frozen_string_literal: true
 
 module WasteExemptionsEngine
-  class RegistrationCompletionService
-    def initialize(transient_registration)
+  class RegistrationCompletionService < BaseService
+    def run(transient_registration:)
```
cintamani: 😍

irisfaraway: Feels good 😎

PR #248, opened 2019-07-31
Pre-populate RenewRegistration with the correct registration data

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-515

It makes sure the transient renew registration is populated with the correct data when a renew registration starts with a correct token.


app/models/concerns/waste_exemptions_engine/can_copy_data_from_registration.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+module WasteExemptionsEngine
+  module CanCopyDataFromRegistration
```
irisfaraway: Yessss 👏

app/models/concerns/waste_exemptions_engine/can_copy_data_from_registration.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    def registration
+      @registration ||= Registration.find_by(reference: reference)
+    end
+
+    private
+
+    def copy_data_from_registration
+      attributes = registration.attributes.except("id",
```
irisfaraway: Does this mean we are copying the old reference to the RenewingRegistration? Is this likely to cause any issues when we complete it?

spec/support/helpers/model_properties.rb
```diff
@@ -51,7 +51,6 @@ module ModelProperties
     TRANSIENT_REGISTRATION_EXEMPTION = (REGISTRATION_EXEMPTION - %i[deregistration_message deregistered_at]).freeze
 
     REGISTRATION = %i[
-      reference
```
irisfaraway: I'm not sure we should be removing this from the attribute list – it means we are no longer testing Registration as completely in other places like https://github.com/DEFRA/waste-exemptions-engine/blob/0b872a0834ced4a95a75b41ea8fbf172e6bc8e82/spec/models/waste_exemptions_engine/registration_spec.rb#L10-L13

app/models/concerns/waste_exemptions_engine/can_copy_data_from_registration.rb
cintamani: No, I have removed the attribute from the one available for copying over :)

cintamani: Is used just for finding it back when we need it :)

PR #253, opened 2019-08-01
Fixes to renew token workflow

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-515

Following QA, fixes included are:
- Redirect to 403 pages so that the correct template is rendered
- Get rid of the temporary `renew` page
- Redirect to the correct transient registration workflow when one is in progress


app/controllers/concerns/waste_exemptions_engine/can_redirect_form_to_correct_path.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module WasteExemptionsEngine
```
Cruikshanks: ❤️ 👍

PR #255, opened 2019-08-01
Set up 'Do you want to renew with these details?' page

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-493

Users should reach this page after clicking the magic link.

To save time, we'll base this on the existing "check your answers" page. However, it does have a different header and different options at the bottom (“Renew with these changes”/”I need to make changes”).


irisfaraway:
Validation message is not customised as there's no way to do that with the gem in its current state. I'll tackle that separately.

PR #256, opened 2019-08-01
Add options to renewal start page

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-493
https://eaflood.atlassian.net/browse/RUBY-498

These buttons will be used to route the user to the declaration or through the longer journey through the forms.

### DIFF
--- a/app/forms/waste_exemptions_engine/renewal_start_form.rb
+++ b/app/forms/waste_exemptions_engine/renewal_start_form.rb
@@ -4,8 +4,19 @@ module WasteExemptionsEngine
   class RenewalStartForm < BaseForm
     attr_accessor :temp_renew_without_changes
 
-    def submit(params)
-      super({}, params[:token])
+    def initialize(registration)
+      super
+      self.temp_renew_without_changes = @transient_registration.temp_renew_without_changes
     end
+
+    def submit(params)
+      # Assign the params for validation and pass them to the BaseForm method for updating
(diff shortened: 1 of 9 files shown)


app/forms/waste_exemptions_engine/renewal_start_form.rb
```diff
@@ -2,8 +2,21 @@
...
+      self.temp_renew_without_changes = @transient_registration.temp_renew_without_changes
+    end
+
     def submit(params)
-      super({}, params[:token])
+      # Assign the params for validation and pass them to the BaseForm method for updating
+      self.temp_renew_without_changes = params[:temp_renew_without_changes]
+      attributes = { temp_renew_without_changes: temp_renew_without_changes }
+
+      super(attributes, params[:token])
```
cintamani: I see that this is a recurring pattern in our forms. And it does feel quite a tangled situation. I have seen the rest of the forms, and with situations like this: https://github.com/DEFRA/waste-exemptions-engine/blob/master/app/forms/waste_exemptions_engine/address_lookup_form.rb#L19:L29 I can see is quite an obliged choice. Maybe we can have some fun at our next dev session and do a 3-peers pairing and try to find a more elegant solution. Would be fun to share some brain power! :D

irisfaraway: Agreed, it is really clunky! Would be great to sort it out.

PR #258, opened 2019-08-02
Use has_secure_token for magic link renew token

cintamani:
We decided that the JWT token is too long and unnecessarly secure for our magic link purposes.
Hence we have decided to use instead an has_secure_token.


app/services/waste_exemptions_engine/generate_renew_token_service.rb
```diff
@@ -1,40 +1,11 @@
...
     def run(registration:)
-      payload = {
-        data: {
-          registration_id: registration.id
-        },
-        exp: exp
-      }
-
-      registration.renew_token = generate_token(payload)
+      registration.regenerate_renew_token
```
cintamani: @irisfaraway I will leave with you the decision of keeping this service or get rid of it. It feels quite unnecessary to me now :D

Cruikshanks: I'd vote for deleting it.

PR #260, opened 2019-08-02
Make it possible to jump directly into the declaration page at renewal

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-498

Add transition events and state to make it possible for the user to go on a different flow to the declaration page when choosing not to make changes to a renewal

### DIFF
--- a/app/views/waste_exemptions_engine/renew_without_changes_forms/new.html.erb
+++ b/app/views/waste_exemptions_engine/renew_without_changes_forms/new.html.erb
@@ -1,0 +1,14 @@
+<%= render("waste_exemptions_engine/shared/back", back_path: back_renew_without_changes_forms_path(@renew_without_changes_form.token)) %>
+
+<div class="text">
+  <%= form_for(@renew_without_changes_form) do |f| %>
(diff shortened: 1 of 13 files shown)


app/views/waste_exemptions_engine/renew_without_changes_forms/new.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= render("waste_exemptions_engine/shared/back", back_path: back_renew_without_changes_forms_path(@renew_without_changes_form.token)) %>
+
+<div class="text">
+  <%= form_for(@renew_without_changes_form) do |f| %>
+    <%= render("waste_exemptions_engine/shared/errors", object: @renew_without_changes_form) %>
+
+    <h1 class="heading-large"><%= t(".heading") %></h1>
+
```
irisfaraway: Are you adding the content for this in a separate PR?

cintamani: Yes :)

PR #263, opened 2019-08-05
Record referring registration when renewing

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-497

When creating a new registration from a renewal, we want to register what registration was renewed.

### DIFF
--- a/spec/models/waste_exemptions_engine/renewing_registration_spec.rb
+++ b/spec/models/waste_exemptions_engine/renewing_registration_spec.rb
@@ -8,6 +8,13 @@ module WasteExemptionsEngine
 
     it_behaves_like "a transient_registration", :renewing_registration
 
+    describe "#registration_attributes" do
+      it "includes a regerral registration id" do
+        attributes = renewing_registration.registration_attributes
+        expect(attributes.keys).to include("referring_registration_id")
+      end
+    end
+
     it "subclasses TransientRegistration" do
       expect(described_class).to be < TransientRegistration
     end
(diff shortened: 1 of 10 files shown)


spec/services/waste_exemptions_engine/registration_completion_service_spec.rb
```diff
@@ -128,6 +128,19 @@ module WasteExemptionsEngine
             expect(ActionMailer::Base.deliveries.count).to eq(old_emails_sent_count + 1)
           end
         end
+
+        context "whne the transient registration is a renewal" do
```
irisfaraway:
```suggestion
        context "when the transient registration is a renewal" do
```

spec/services/waste_exemptions_engine/registration_completion_service_spec.rb
```diff
@@ -128,6 +128,19 @@ module WasteExemptionsEngine
...
+          let(:renewing_registration) { create(:renewing_registration) }
+
+          it "copy over data about referring registration" do
```
irisfaraway:
```suggestion
          it "copes over data about the referring registration" do
```

spec/models/waste_exemptions_engine/renewing_registration_spec.rb
```diff
@@ -8,6 +8,13 @@ module WasteExemptionsEngine
 
     it_behaves_like "a transient_registration", :renewing_registration
 
+    describe "#registration_attributes" do
+      it "includes a regerral registration id" do
```
irisfaraway:
```suggestion
      it "includes a referring registration id" do
```

PR #266, opened 2019-08-06
Implement design of Renew completed page

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-500

Implement design of renew completed page.

Screenshot 2019-08-06 at 12 17 07


app/forms/waste_exemptions_engine/renewal_complete_form.rb
```diff
@@ -1,21 +1,7 @@
...
-
-    # Override BaseForm method as users shouldn't be able to submit this form
-    def submit(_params)
-      raise UnsubmittableForm
-    end
-
-    def self.can_navigate_flexibly?
-      false
-    end
+  class RenewalCompleteForm < RegistrationCompleteForm
```
irisfaraway: If we are trying to share some code, rather than subclassing a concrete form, I'd prefer it if we had some sort of `CompleteForm` that both of these forms used. We do have an existing pattern of using modules for this sort of thing – have a look at `AddressForm` and `DataOverviewForm`.

cintamani: I want to implement this first: https://github.com/DEFRA/ruby-services-team/issues/59 and then we can look at this. The main "issue" is that the edit competition form will not make use of it yet. We might find a different way using composition rather than inheritance after we have dealt with the issue above :)

irisfaraway: In that case, I'd rather have the duplication now rather than introduce subclassing one concrete form from another – let's go for consistency with the other forms and then it'll be easier to clean it all up later.

config/initializers/time_formats.rb
```diff
@@ -4,3 +4,5 @@
 Date::DATE_FORMATS[:day_month_year] = "%-d %B %Y"
 # Returns a time in this format: 1:23am on 1 January 2019
 Time::DATE_FORMATS[:time_on_day_month_year] = "%-l:%M%P on %-d %B %Y"
+# Returns a date in this format: January 2019
+Date::DATE_FORMATS[:time_on_month_year] = "%B %Y"
```
Cruikshanks: It's picky, but as time does not form part of this format I'd look to rename this

```suggestion
Date::DATE_FORMATS[:month_year] = "%B %Y"
```

app/controllers/waste_exemptions_engine/renewal_complete_forms_controller.rb
```diff
@@ -6,7 +6,11 @@ def new
       return unless super(RenewalCompleteForm, "renewal_complete_form")
 
       registration = RegistrationCompletionService.run(transient_registration: @transient_registration)
+
       @renewal_complete_form.reference = registration.reference
+
+      registration_expires_on = registration.registration_exemptions.first.expires_on
+      @renewal_complete_form.expire_month_year = registration_expires_on.to_formatted_s(:time_on_month_year)
```
cintamani:
```suggestion
      @renewal_complete_form.expire_month_year = registration_expires_on.to_formatted_s(:month_year)
```

Cruikshanks:
I've marked the 2 issues raised by CodeClimate as `WONTFIX` for now @cintamani , as we intend to look at how we reuse code across the forms at a later date.

Perfect example of what we discussed in our retro yesterday, of putting something on credit which we need to pay back later 😁 !

PR #274, opened 2019-08-09
Display error when trying to renew too late

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-496

We have certain business rules for when a registration can and cannot be renewed.

This includes a 'grace period' after the registration expires when it can still be renewed. This is 30 days after the expiry date.

If an expired registration is no longer in the grace period, it should not be possible to renew it. Anyone trying to access the renewal at any point in its process (clicking the magic link, loading a page midway through the renewal process) should see a useful page explaining that the registration can no longer be renewed.


irisfaraway:
Will make another change to this once https://github.com/DEFRA/waste-exemptions-engine/pull/276 is merged so I'll get rid of the review requests for now.

irisfaraway:
@cintamani Aha, good point. I agree we should move them into the engine and remove them from the back office, so I'll get on with that 👍

spec/dummy/config/initializers/waste_exemptions_engine.rb
```diff
@@ -20,4 +20,8 @@
 
   # Last Email Cache config
   config.use_last_email_cache = ENV["USE_LAST_EMAIL_CACHE"] || "false"
+
+  # Renewing config
+  config.renewal_window_open_before_days = ENV["RENEWAL_WINDOW_OPEN_BEFORE_DAYS"] || 28
```
Cruikshanks: My only quibble is around the env var names, which then has had a knock on effect to the config and other names.

They are both doing the same thing, holding the number of days for a named period of time. However we seem to have ended up with a different naming convention for both.

My preference would be

```suggestion
  config.registration_renewal_window = ENV["REGISTRATION_RENEWAL_WINDOW"] || 28
  config.registration_grace_window = ENV["REGISTRATION_GRACE_WINDOW"] || 30
```

I've never been that bothered about stating it's in days, but I know @cintamani has had other thoughts on this. But my main aim is to get the names consistent. So as long as that is achieved I'll be happy 😁

cintamani: Eheh :D I wanted to make sure the ENV variable contains an integer that express Days, so I am keep to keep the `-days` stated somewhere for documentation purposes. Just as good practice that any variable states what kind of value it holds, really :)

irisfaraway: How do we all feel about `renewal_window_before_expiry_in_days` and `renewal_window_after_expiry_in_days`? Feels like the most explicit about what it is.

PR #284, opened 2019-08-12
Set up custom dimensions to track workflows

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-509

We want to be able to track the performance of the new registration journey and the renewals journey. Because they use the same pages, we'll need a different way to differentiate.

This PR implements custom dimensions in the data layer to be used by Google Tag Manager and Google Analytics. This will allow us to filter pageviews by the journey, and will potentially let us do even more with the data layer in the future.


irisfaraway:
Decided there's no reason to have this in the engine as we're only after front office analytics. Closing and will reimplement in the front office.

PR #286, opened 2019-08-14
[WIP] Add ActiveSupport::Callbacks on BaseForm `#initialize`

cintamani:
This is step one in the direction of forms refactoring.
It adds support for callbacks in the `BaseForm#initialize` method and changes all forms to use callbacks instead of overriding the method itself.

Part of: https://github.com/DEFRA/ruby-services-team/issues/59


cintamani:
Putting this back in WIP as I realised there is no test coverage for the attribute setters in initialize.

cintamani:
Test coverage added in: https://github.com/DEFRA/waste-exemptions-engine/pull/287

Cruikshanks:
I've not reviewed everyline of code but have focused on the new concerns, and the use of callbacks instead of `initialize()` and `super()`.

I took a simple form to start with (the `applicant_email_form.rb`) and compared the changes made to what we already had. TBH, I was initially concerned. The form in its current (live) state was pretty simple, and what was going on when initialized was easy to follow.

```ruby
# applicant_email_form
def initialize(registration)
  super
  self.applicant_email = @transient_registration.applicant_email
  self.confirmed_email = @transient_registration.applicant_email
end

# base_form
def initialize(registration)
  # Get values from registration so form will be pre-filled
  @transient_registration = registration
  self.token = @transient_registration.token
end
```

To me, that's simples! Compare that with the new code

```ruby
# applicant_email_form
include CanSetApplicantEmail

set_callback :initialize, :after, :set_confirmed_email

private

def set_confirmed_email
  self.confirmed_email = @transient_registration.applicant_email
end

# can_set_applicant_email
included do
  attr_accessor :applicant_email

  set_callback :initialize, :after, :set_applicant_email

  def set_applicant_email
    self.applicant_email = @transient_registration.applicant_email
  end
end

# base_form
include ActiveSupport::Callbacks
define_callbacks :initialize

def initialize(registration)
  run_callbacks :initialize do
    # Get values from registration so form will be pre-filled
    @transient_registration = registration
    self.token = @transient_registration.token
  end
end
```

There feels to me a lot more complexity in getting the form initialized. I saw the light though when I looked at the `registration_complete` and `renewal_complete` forms and the changes that had been made there.

```ruby
# old code
attr_accessor :reference, :exemptions_plural, :applicant_email, :contact_email, :emails_plural

def initialize(registration)
  super
  self.reference = @transient_registration.reference
  self.exemptions_plural = @transient_registration.exemptions.length > 1 ? "many" : "one"
  self.applicant_email = @transient_registration.applicant_email
  self.contact_email = @transient_registration.contact_email
  self.emails_plural = applicant_email == contact_email ? "one" : "many"
end

# new code
include CanSetApplicantEmail
include CanSetContactEmail
include CanSetReference
include CanSetExemptionsAndEmailsPlural
```

Lots of lovely reuse and removal of duplicate code ❤️😊. So I still have some concerns about someone new to the code being able to follow things through, but at this time my 'spidey sense' is telling me once you are over that hurdle things will actually be cleaner and simpler to follow in the long run.

So am happy to keep going down this path. Would love to get @irisfaraway take on this as well though.

irisfaraway:
@Cruikshanks I think I had the same initial concerns about it being less immediately clear to someone who is new to the code. But the reduction in duplication is cool, and I think these changes will set us up for even more refactoring and code reuse in the future too.

To help with this, I think it would be good to get some documentation going about how the form objects work, including some info about these new concerns. I'd suggest we add an issue to deal with this.

cintamani:
Thanks for your feedback guys! I am still thinking about this, mostly the part about setting values. 
I also want to point you to: https://github.com/DEFRA/waste-exemptions-engine/pull/286/files#diff-d929aeae98d184705f0ec61f960e4b6fL15 
Which is the actual case for which the hooks are so cool and reduce complexity. That is the ACTUAL case for them to be useful. 

BUT, in regard of the attribute setters, I agree with you it doesn't really look that good, which is why I have putted this on hold and worked on the submit. 

The thing is that we are setting params on the form twice, once in the `initialize`, getting them from the `transient_registration` (which is already a class instance variable, so we could easily delegate)  , and then again before submitting, for the sake of validation and update of the transient registration. 
I was thinking to try solving the problem with something like: 
```
class SomeForm 
  attr_accessor :transient_registration

  def initialize(transient_registration) 
    @transient_registration = transient_registration
  end
 
 # Attribute getting method, rather than instance variable setting
  def postcode
    @postcode ||= transient_registration.postcode
  end

  # Setting value from submit:
  def set_form_parameters(params)
    @postcode = params[:postcode]
  end
end
```

What do you think about that approach?

irisfaraway:
@cintamani I think that sounds like a good start :) Could you do a proof of concept with one of the forms, maybe in a different branch?

cintamani:
@irisfaraway Yes! will do that :)

cintamani:
@irisfaraway https://github.com/DEFRA/waste-exemptions-engine/compare/proof-of-concept-refactoring-forms?expand=1 This is what I mean :)

Cruikshanks:
> @irisfaraway https://github.com/DEFRA/waste-exemptions-engine/compare/proof-of-concept-refactoring-forms?expand=1 This is what I mean :)

In that example would we still be able to move `first_name` and `last_name` into a concern?

cintamani:
@Cruikshanks depends on the name of the variable in the `transient_registration`. When that matches, yes. But, do you think we should? Sometimes they add complexity and reduce readability rather than help, and this might be one of those cases

irisfaraway:
@cintamani I'm wondering if we could move even more logic out of `submit`. Quick example:

```
    def submit(params)
      attributes = assign_params_and_attributes_for_submit(params)
      super(attributes)
  end

  private

  def assign_params_and_attributes_for_submit
      @first_name = params[:first_name]
      @last_name = params[:last_name]

      attributes = {
        applicant_first_name: first_name,
        applicant_last_name: last_name
      }
  end
```

Cruikshanks:
I might be getting out of my depth now, but I've combined what @cintamani was suggesting with what you suggested @irisfaraway , with my own twist on the attributes to support reuse.

https://github.com/DEFRA/waste-exemptions-engine/compare/proof-of-concept-refactoring-forms-2?expand=1

Possibly my commit title is apt 😁

cintamani:
@irisfaraway exactly what I am suggesting too with https://github.com/DEFRA/waste-exemptions-engine/compare/submit-refactoring-off-setters?expand=1 we are definitely getting on the same page!! Love this! 
@Cruikshanks I like your suggestion too, but I was going for splitting the assignment of attributes from the hash of registration attributes in 2 methods rather than one :) and in regard of concerns, are you sure? I am still wondering if is a good idea to externalise those bits there or leave them in the form classes. Feels like leaving them will improve readability. It might be OK to have concerns for stuff like `postcode` since we have many more than only 2 forms using that. What do you think?

Cruikshanks:
@cintamani I taken my example one step further https://github.com/DEFRA/waste-exemptions-engine/compare/proof-of-concept-refactoring-forms-2?expand=1

TL;DR you are right about readability, but there is repetition which the concerns can remove. And that means there is *a lot* of code we can delete. Also I'd happily do the attribute assignment across 2 methods rather than one, but I'm not sure how to make it work in my example so have left as is for now.

PR #289, opened 2019-08-15
Add settings to use parallel testing on Travis

cintamani:
This adds all necessary config to use the parallel tests gem on travis and speed up the run time of our test suite CI.

It adds parallelism of 4 and gains 2 minutes less of running time on travis. From the current ~6 minutes, to an average of 4 minutes. 

Closes: https://github.com/DEFRA/ruby-services-team/issues/61


spec/dummy/config/database.yml
```diff
@@ -15,7 +15,7 @@ development:
 
 test:
   <<: *default
-  database: waste_exemptions_test
+  database: waste_exemptions_test_<%= ENV["TEST_ENV_NUMBER"] %>
```
cintamani:
```suggestion
  database: waste_exemptions_test
```

Rakefile
```diff
@@ -1,5 +1,11 @@
 # frozen_string_literal: true
 
+begin
+  require "parallel_tests/tasks"
+rescue LoadError
+  puts "No parallel_tests available. Perhaps running in production mode?"
```
Cruikshanks: I'd rather we just output nothing if this happens, like we do in the rspec block below.

It shouldn't impact anything but we are reliant on rake being run in production mode in a number of different scenarios, such as when we deploy, or when we run automated jobs in production.

If we are running parallel tests and the require fails, we'll know about it anyway when the next step fails. So as rake is more often used in a production mode, I'd prefer not have this message show each time we do.

cintamani:
```suggestion
  # No parallel_tests available. Perhaps running in production mode. 
```

Rakefile
```diff
@@ -1,5 +1,11 @@
 # frozen_string_literal: true
 
+begin
+  require "parallel_tests/tasks"
+rescue LoadError
+  # No parallel_tests available. Perhaps running in production mode.
```
cintamani:
```suggestion
  # No parallel_tests available. Perhaps running in production mode.
```

PR #293, opened 2019-08-16
Use validation gem for business type

irisfaraway:
We now have a business type validator built into defra-ruby-validators, so we can use that instead of having our own validator in the engine.


Cruikshanks:
Oooh!
excited

irisfaraway:
@Cruikshanks Many more to come! https://media2.giphy.com/media/rVbAzUUSUC6dO/giphy.gif

PR #305, opened 2019-08-20
For discussion on Thursday

cintamani:
This is a `template` of the fixes I was thinking to introduce.
Basically, we are keeping the `form` pattern main methods in the Parent classes, and the children define the rest of specific things but without the re-write of the main methods. 
Instead, the parent class offers a `default` simple behaviour that works on the simplest cases and the children will deal with the parameters complications. 
A `delegate` is used instead of assigning instance variable on the form object, and at validation time, we assign values to the registration itself before validating, as we would do with a controller. If valid we save, otherwise, we don't. 
I think this narrows down the form's responsibilities which right now are too many.

### DIFF
--- a/app/forms/waste_exemptions_engine/base_form.rb
+++ b/app/forms/waste_exemptions_engine/base_form.rb
@@ -5,6 +5,15 @@ module WasteExemptionsEngine
     include ActiveModel::Model
     include CanStripWhitespace
 
+    attr_reader :transient_registration
+
+    delegate :token, to: :transient_registration
+
+    # If the record is new, and not yet persisted (which it is when the start
+    # page is first submitted) then we have nothing to validate hence the check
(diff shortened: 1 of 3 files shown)


app/forms/waste_exemptions_engine/base_form.rb
```diff
@@ -17,38 +26,35 @@ def self.can_navigate_flexibly?
...
+    end
 
     private
 
+    def transient_registration_attributes(_params)
+      # Default behaviour. Subclasses to override.
+      {}
+    end
+
     def registration_valid?
```
cintamani: Do we really need this? I don't think we have any validations in place yet on the registration itself that is not re-defined in thee form. Let's discuss the possibility to drop this code.

Cruikshanks:
excited

❤️ 🎉

PR #310, opened 2019-08-27
Redirect to start form page when a token is invalid

cintamani:
Part of: https://eaflood.atlassian.net/browse/RUBY-603

When a user hits a form page with an invalid token, the current code will generate a `NewRegistration`, assign it the workflow state for the form the user has requested and try to render the page. The problem is that if a registration is created in this way, whithout ever going through the first `start_form` submit, that registration has no token assigned to it. Hence, when the view renders, it will generate an error as the back button url cannot be generated.

This change will make sure that when a token is invalid and hence a `NewRegistration` is created and not persisted, its status is not changed at all from the default `start_form` and so the user is redirected to the start form page.


app/controllers/waste_exemptions_engine/forms_controller.rb
```diff
@@ -67,12 +67,9 @@ def setup_checks_pass?
     def set_workflow_state
       return unless state_can_navigate_flexibly?(@transient_registration.workflow_state)
       return unless state_can_navigate_flexibly?(requested_state)
+      return unless @transient_registration.persisted?
 
-      if @transient_registration.persisted?
-        update_workflow_state(requested_state)
-      else
-        assign_workflow_state(requested_state)
```
irisfaraway: Does this mean we can delete `def assign_workflow_state`?

cintamani: Oh yeah we can!

PR #315, opened 2019-08-28
Use applicant first and last name transient registration params

cintamani:
To facilitate the form refactoring and take advantage of delegate, we want to use transient registration attributes names were possible

### DIFF
--- a/app/views/waste_exemptions_engine/applicant_name_forms/new.html.erb
+++ b/app/views/waste_exemptions_engine/applicant_name_forms/new.html.erb
@@ -19,14 +19,14 @@
       </fieldset>
     </div>
 
-    <div class="form-group <%= "form-group-error" if @applicant_name_form.errors[:last_name].any? %>">
-      <fieldset id="last_name">
-        <% if @applicant_name_form.errors[:last_name].any? %>
-          <span class="error-message"><%= @applicant_name_form.errors[:last_name].join(", ") %></span>
+    <div class="form-group <%= "form-group-error" if @applicant_name_form.errors[:applicant_last_name].any? %>">
+      <fieldset id="applicant_last_name">
+        <% if @applicant_name_form.errors[:applicant_last_name].any? %>
+          <span class="error-message"><%= @applicant_name_form.errors[:applicant_last_name].join(", ") %></span>
         <% end %>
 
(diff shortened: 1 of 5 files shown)


app/views/waste_exemptions_engine/applicant_name_forms/new.html.erb
```diff
@@ -8,25 +8,25 @@
...
-            <span class="error-message"><%= @applicant_name_form.errors[:first_name].join(", ") %></span>
+    <div class="form-group <%= "form-group-error" if @applicant_name_form.errors[:applicant_first_name].any? %>">
+      <fieldset id="applicant_first_name">
+        <% if @applicant_name_form.errors[:applicant_first_name].any? %>
+            <span class="error-message"><%= @applicant_name_form.errors[:applicant_first_name].join(", ") %></span>
         <% end %>
 
-        <%= f.label :first_name, t(".first_name_label"), class: "form-label" %>
-        <%= f.text_field :first_name, value: @applicant_name_form.first_name, class: "form-control govuk-input--width-20" %>
+        <%= f.label :applicant_first_name, t(".first_name_label"), class: "form-label" %>
```
irisfaraway: Can we update the I18n bits while we're at it? (eg `.applicant_first_name_label`) Will just bring everything into line then :)

cintamani: Sounds good :)

spec/forms/waste_exemptions_engine/applicant_name_form_spec.rb
```diff
@@ -10,27 +10,27 @@ module WasteExemptionsEngine
...
+        expect(validators.keys).to include(:applicant_first_name)
+        expect(validators[:applicant_first_name].first.class)
           .to eq(WasteExemptionsEngine::PersonNameValidator)
       end
 
       it "validates the last name using the PersonNameValidator class" do
-        expect(validators.keys).to include(:last_name)
-        expect(validators[:first_name].first.class)
+        expect(validators.keys).to include(:applicant_last_name)
+        expect(validators[:applicant_first_name].first.class)
```
irisfaraway:
```suggestion
        expect(validators[:applicant_last_name].first.class)
```

Old bug, but just spotted this!

PR #318, opened 2019-08-28
Make confirmed email an attribute of the form

cintamani:
As part of the form refactoring, we want to be able to use delegate on our attributes and not re-write main initialize and submit methods.
In the case of the applicant confirmed email, we want the form to deal with the attribute in this different way, were we are able to set and read it and overwrite its value.
When sorting out the `submit` method, we will also change the setting of the attribute from params making use of callbacks with parameters.

### DIFF
--- a/app/forms/waste_exemptions_engine/applicant_email_form.rb
+++ b/app/forms/waste_exemptions_engine/applicant_email_form.rb
@@ -2,7 +2,8 @@
 
 module WasteExemptionsEngine
   class ApplicantEmailForm < BaseForm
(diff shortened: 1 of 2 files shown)


cintamani:
I tend to agree with the first statement too, I only have doubt in this case because of the nature of this particular attribute. I suggest we keep going with the refactoring keeping the parameters in the Forms, and that's because the Forms holds the validations, and then come back to it after the refactoring in done, and see if is better to move them to the model instead. The thing is that right now I don't know how many other instances of this we will have to maintain. WDYT?

PR #324, opened 2019-08-28
Forms refactoring - Use helpers to format names

cintamani:
Instead of adding an attr_accessor to the form in order to format data for the view, we have decided to use Helpers.
This replace the form-defined methods with a `format_names` helper method used for applicant and contact names.


app/helpers/waste_exemptions_engine/application_helper.rb
```diff
@@ -9,6 +9,10 @@ def title
       title_elements.join(" - ")
     end
 
+    def format_names(first_name, last_name)
```
irisfaraway: Can we have a unit test for this?

PR #325, opened 2019-08-29
Forms refactoring - Exemptions form division of responsibilities

cintamani:
Leave to the view what belongs to the view and to the form what belongs to the form :D

This change remove the list of exemptions being assigned to the form using the attribute name that is suppose to return the transient registration exemptions. It instead uses it as the real list of exemptions that are part of a registration or assign to them the registration ticked in the form itself. An helper is instead responsible to fetch the list of all exemptions for the view to show.

### DIFF
--- a/app/helpers/waste_exemptions_engine/exemptions_forms_helper.rb
+++ b/app/helpers/waste_exemptions_engine/exemptions_forms_helper.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+module WasteExemptionsEngine
+  module ExemptionsFormsHelper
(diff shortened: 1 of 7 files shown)


app/helpers/waste_exemptions_engine/exemptions_forms_helper.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module WasteExemptionsEngine
```
Cruikshanks: Just a request for a unit test for this 😁 !

PR #327, opened 2019-08-30
Move view logic away from forms - Renewal and Registration complete

cintamani:
This removes the `_plural` methods away from the renewals and registration forms and move them to helpers.


cintamani:
I think it makes more sense. Will do that :) thanks

PR #339, opened 2019-09-03
Do not save value of :address_finder_error at initialize

cintamani:
The main reason why I am removing this, is because I believe it makes no difference to the user experience.
I could not find any issue caused by this removal and so I am opening a separate PR for this change so we can decide if we can remove it or not. I miss some history :D


cintamani:
And.... I figured 🤦‍♂

PR #340, opened 2019-09-04
Add transient addresses has_one associations to TransientRegistration

cintamani:
With this little trick, we don't have to worry any more about stuff like "Can add or replace an address" as we have the ability to manipulate the transient addresses separate.
It does also make more sense logically and makes the form nested attribute implementation much easier.

The code using this is part of a bigger PR that will land later.

### DIFF
--- a/app/models/waste_exemptions_engine/transient_registration.rb
+++ b/app/models/waste_exemptions_engine/transient_registration.rb
@@ -17,6 +17,15 @@ module WasteExemptionsEngine
     validates_presence_of :token, on: :save
 
     has_many :transient_addresses, dependent: :destroy
+
+    # In this case, we have to pass the correct enum value, as `enum` will not generate the right query in this case.
+    has_one :transient_site_address, -> { where(address_type: 3) }, class_name: "TransientAddress"
+    has_one :transient_contact_address, -> { where(address_type: 2) }, class_name: "TransientAddress"
+    has_one :transient_operator_address, -> { where(address_type: 1) }, class_name: "TransientAddress"
+    accepts_nested_attributes_for :transient_site_address
+    accepts_nested_attributes_for :transient_contact_address
+    accepts_nested_attributes_for :transient_operator_address
+
     has_many :transient_people, dependent: :destroy
     has_many :transient_registration_exemptions, dependent: :destroy
     has_many :exemptions, through: :transient_registration_exemptions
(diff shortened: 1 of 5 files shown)


cintamani:
@irisfaraway allow for `accept_nested_parameters` to work. The difference is that it returns an object like ActiveRecordRelation were you can chain rails methods while the implementation of `#find_address_by_type` will return a `TransientAddress` object directly

cintamani:
Also, you are able to assign a `transient_site_address`. You can do stuff like `transient_registration.build_transient_site_address` directly, and it will replace the one in the `transient addresses` array, so we don't need to do stuff like https://github.com/DEFRA/waste-exemptions-engine/blob/master/app/forms/concerns/waste_exemptions_engine/can_add_or_replace_an_address.rb#L8

Basically, allow to use rails to do rails stuff XD

irisfaraway:
@cintamani Would it make sense to replace the existing methods with something like this then? (For registrations and their addresses as well).

I'm just wondering if having two very similar methods for different purposes (`transient_registration.transient_site_address` and `transient_registration.site_address`) could cause confusion.

cintamani:
@irisfaraway I hope to get rid of these: https://github.com/DEFRA/waste-exemptions-engine/blob/e3f715f5bf6e4753cf243651ca6d909fa1b77fe7/app/models/concerns/waste_exemptions_engine/can_have_registration_attributes.rb#L26-L36 when I have done the other forms too :)

irisfaraway:
@cintamani What happens if we remove the `transient_` bits from the names of the new relationships? Can we just do an immediate replace of the old methods as part of this PR?

I think they are used a fair amount in the back office as well so it'd save a lot of find-and-replacing.

cintamani:
@irisfaraway I could do that, but I find it very confusing that they don't include the `transient_` bit. They confuse me as I always expect a `Address` object and I get a `TransientAddress` one, so I prefer to deal with the find and replace :D

irisfaraway:
@cintamani Ahh, I prefer not having `transient` as I think it clutters up the code and assume that the sub-relations of a `transient_registration` will always be `transient_` things anyway. So maybe we need @Cruikshanks for the deciding vote!

irisfaraway:
Also then we can call `.site_address` on an object without worrying whether it is a `Registration` or `TransientRegistration`, which is useful for a lot of things

cintamani:
Yep found them! Will `:alias` :)

cintamani:
As per face-to-face discussion, I am going to merge this and open a different PR for adding :alias and removing the implementation of the other methods

PR #341, opened 2019-09-04
Refactor Contact Address Manual form

cintamani:
This removes inheritance just so the class is in a clear state and so I can work on the different address form separately. Once the refactor of the BaseForm is done, will be easier to put together common bits of code (With CodeClimates help too!).
The parameters, in this case, are coming through the StrongParameter Rails implementation.
This allow for `accept_nested_attributes` to work, which will then allow us to implement the `delegate` approach to the forms.
The object passed to the shared partial is also changed, so that we receive the parameters correctly nested.
Everything else is fixes around that.
Particularly the specs were to change so that they will pass over proper params object. This do not affect existing forms still dealing with hashes, as the params object does behave as such already.
We already agreed that strong parameters will be the way to go, and once that's completed then we can move the implementation away from the forms and back into the controller.


app/validators/waste_exemptions_engine/legacy_manual_address_validator.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module WasteExemptionsEngine
+  class LegacyManualAddressValidator < ActiveModel::Validator
```
cintamani: This is here for convenience. Will be removed once this refactoring is applied to the other manual address forms.

app/views/waste_exemptions_engine/contact_address_manual_forms/new.html.erb
```diff
@@ -19,7 +19,9 @@
       <%= link_to(t(".postcode_change_link"), back_contact_address_manual_forms_path(@contact_address_manual_form.token)) %>
     </div>
 
-    <%= render("waste_exemptions_engine/shared/manual_address", form: @contact_address_manual_form, f: f) %>
+    <%= f.fields_for :transient_contact_address do |f| %>
```
cintamani: This allow Rails to generate inputs with the correct name, so that the parameters sent via form submit will arrive in the correct nested state.

spec/dummy/config/locales/manual_address_validator/en.yml
```diff
@@ -17,3 +17,18 @@ en:
               too_long: "The town or city must be no longer than 30 characters"
             postcode:
               too_long: "The postcode must be no longer than 8 characters"
+        test/legacy_manual_address_validatable:
```
cintamani: Ditto of https://github.com/DEFRA/waste-exemptions-engine/pull/341/files#r320812535

spec/forms/waste_exemptions_engine/contact_address_manual_form_spec.rb
```diff
@@ -4,13 +4,130 @@
 
 module WasteExemptionsEngine
   RSpec.describe ContactAddressManualForm, type: :model do
-    it_behaves_like "a manual address form", :contact_address_manual_form
```
cintamani: The code in this spec will be need to be moved back in a shared example once the rest of the forms are in the same state.

spec/forms/waste_exemptions_engine/contact_address_manual_form_spec.rb
```diff
@@ -4,13 +4,130 @@
...
+              premises: "",
+              street_address: "",
+              locality: "",
+              city: "",
+              postcode: ""
+            }
+          },
+          {
+            token: form.token,
+            transient_contact_address: {
```
cintamani: This is an example of how we will receive the params now

app/forms/waste_exemptions_engine/contact_address_manual_form.rb
```diff
@@ -1,8 +1,63 @@
...
+      self.address_finder_error = @transient_registration.address_finder_error
+      transient_registration.update_attributes(address_finder_error: nil)
+
+      # Prefill the existing address unless the postcode has changed from the existing address's postcode
+      prefill_transient_contact_address_params if saved_address_still_valid?
+    end
+
+    def submit(params)
+      permitted_attributes = params.require(:transient_contact_address)
+      permitted_attributes = permitted_attributes.permit(:locality, :postcode, :city, :premises, :street_address, :mode)
```
cintamani: This will be moved to the controller later :)

app/forms/waste_exemptions_engine/contact_address_manual_form.rb
```diff
@@ -1,8 +1,63 @@
...
+
+      # Needed for validation
+      assign_params(permitted_attributes)
+
+      super(transient_contact_address_attributes: permitted_attributes)
+    end
+
+    private
+
+    def assign_params(permitted_attributes)
```
cintamani: all this will disappear thanks to delegate later :)

app/validators/waste_exemptions_engine/manual_address_validator.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module WasteExemptionsEngine
-  class ManualAddressValidator < ActiveModel::Validator
+  class ManualAddressValidator < ActiveModel::EachValidator
```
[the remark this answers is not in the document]
cintamani: Won't fix

app/validators/waste_exemptions_engine/legacy_manual_address_validator.rb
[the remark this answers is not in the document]
cintamani: Won't fix

app/views/waste_exemptions_engine/contact_address_manual_forms/new.html.erb
```diff
@@ -19,7 +19,9 @@
       <%= link_to(t(".postcode_change_link"), back_contact_address_manual_forms_path(@contact_address_manual_form.token)) %>
     </div>
 
-    <%= render("waste_exemptions_engine/shared/manual_address", form: @contact_address_manual_form, f: f) %>
+    <%= f.fields_for :contact_address do |f| %>
```
cintamani: This allow Rails to generate inputs with the correct name, so that the parameters sent via form submit will arrive in the correct nested state.

PR #343, opened 2019-09-05
Use rails associations to define associations with the addresses

cintamani:
In both the Registration and Transient registration we are implementing Rails association in order to associate those registration with the correct addresses.
We have decided that the name of the methods should be the same for bot Transient registration and non transient registration.
This allow for us to be able to easely reuse the same code on the 2 kind of object, and at the same time gives us the ability to edit and update the addresses on those registration without having to deal with the addresses array association complication.

### DIFF
--- a/spec/factories/edit_registration.rb
+++ b/spec/factories/edit_registration.rb
@@ -16,7 +16,6 @@ FactoryBot.define do
     trait :modified do
       after(:build) do |edit_registration|
         registration = edit_registration.registration
-        # binding.pry
         # Update string attributes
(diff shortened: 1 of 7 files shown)


cintamani:
Yep! I forgot to check for that :D

spec/factories/edit_registration.rb
```diff
@@ -16,7 +16,7 @@
     trait :modified do
       after(:build) do |edit_registration|
         registration = edit_registration.registration
-
+        # binding.pry
```
cintamani:
```suggestion
```

PR #347, opened 2019-09-11
Identify EA Administrative Area for new regs.

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-483

The previous WEX service would attempt to identify the EA Admin area for a site once it had been entered as part of the registration. This stopped working in December 2018.

When we shipped the new version of the service we left out this functionality because

- we were still awaiting a fix
- we knew both FRAE and WEX depend on this so wanted to implement a shared solution

With the creation of [defra-ruby-area](https://github.com/DEFRA/defra-ruby-area) we now have that shared solution.

So this change covers everything needed to integrate the gem and identify the area as part of the new registration journey.

### DIFF
--- a/app/services/waste_exemptions_engine/area_lookup_service.rb
+++ b/app/services/waste_exemptions_engine/area_lookup_service.rb
@@ -31,7 +31,7 @@ module WasteExemptionsEngine
 
     def handle_error(error)
       Airbrake.notify(error, easting: @easting, northing: @northing) if defined? Airbrake
-      Rails.logger.error "Area lookup failed: #{error}"
+      Rails.logger.error "Area lookup failed:\n #{error}"
     end
 
   end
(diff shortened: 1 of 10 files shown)


app/services/waste_exemptions_engine/area_lookup_service.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    def valid_arguments?
+      return false unless @easting.is_a?(Numeric) && @northing.is_a?(Numeric)
+
+      true
+    end
+
+    def handle_error(error)
+      Airbrake.notify(error, easting: @easting, northing: @northing) if defined? Airbrake
+      Rails.logger.error "Area lookup failed: #{error}"
```
cintamani:
```suggestion
      Rails.logger.error "Area lookup failed:\n #{error}"
```

PR #364, opened 2019-09-25
Do not run area lookups on 0 coordinates

cintamani:


cintamani:
Codeclimate is a won't fix

PR #365, opened 2019-09-26
Restrict address missing scopes to site addresses

Cruikshanks:
The various scopes on Address to identify those with missing x & y, and area information is used in jobs to ensure they are populated. The aim is to try and identify the EA area for each registration we have.

The address we use to determine this though is the **site**. Where the site is located is what determines the a registration's area. So we don't care if the x & y or area is missing on the other addresses. We only care they are populated for the site address.

### DIFF
--- a/spec/models/waste_exemptions_engine/address_spec.rb
+++ b/spec/models/waste_exemptions_engine/address_spec.rb
@@ -40,6 +40,9 @@ module WasteExemptionsEngine
       describe ".sites_with_easting_and_northing" do
         it "returns all site addresses with x and y information" do
           create(:address, x: nil, y: 123.4)
+          create(:address, x: 123.4, y: nil)
+          create(:address, x: 123.4, y: 123.4)
+          create(:address, :site_address, x: nil, y: 123.4)
           create(:address, :site_address, x: 123.4, y: nil)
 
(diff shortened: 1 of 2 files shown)


spec/models/waste_exemptions_engine/address_spec.rb
```diff
@@ -18,45 +18,50 @@ module WasteExemptionsEngine
...
+          expect(described_class.sites_missing_easting_or_northing).to match_array(missing_info_records)
         end
       end
-      describe ".with_easting_and_northing" do
-        it "returns all address with x and y information" do
+      describe ".sites_with_easting_and_northing" do
+        it "returns all site addresses with x and y information" do
           create(:address, x: nil, y: 123.4)
-          create(:address, x: 123.4, y: nil)
+          create(:address, :site_address, x: 123.4, y: nil)
```
andrewhick: Not a showstopper, but for completeness could we also have:
- a non-site address missing a northing
- a site address missing an easting?

Cruikshanks: stormoff

andrewhick: You've gone above and beyond 👍

PR #367, opened 2019-09-27
Operator Address manual form refactoring

cintamani:
Following on: https://github.com/DEFRA/waste-exemptions-engine/pull/341/files
This apply the same refactoing to the operator address manual form.
The goal is to make use of rails-ness on these forms, so we want them to name the attributes as per TransientAddress attributes. This will allow the use of strong parameters and less code to custom-submit these forms rather than letting rails do the job for us :D


cintamani:
Keep ignoring CodeClimate for now

PR #375, opened 2019-09-30
Refactor address lookups form

cintamani:
Part of the form refactoring.
In this case I have kept the base Lookup Forms class, but I have moved the implementation of the submit method in the children.
This will allow easier refactoring for the rest of the class.
The create address from urpn function have moved to a concern and has no more any dependency with methods defined in the including classes.
Use delegate and alias instead of attribute accessors.


cintamani:
Code climate is a won't fix.

PR #383, opened 2019-10-04
Forms refactoring: Submit and Strong parameters

cintamani:
Forms refactoring: get rid of `submit` re-implementation in standard forms and instead use rails strong parameters. 
By implementing it we also have simplified our lookup forms by only passing `uprn` as a normal paramter, and the rest of lookups now happen in an active record hook.


cintamani:
CodeClimate is all a `won't fix`

app/forms/waste_exemptions_engine/address_lookup_form_base.rb
```diff
@@ -21,19 +18,16 @@ def look_up_addresses
...
-      return false unless existing_address
-      return false unless existing_address.uprn.present?
+      data = temp_addresses.detect { |address| address["uprn"] == uprn.to_i }
+      return {} unless data
 
-      true
+      data
+        .except("address")
+        .except("state_date")
+        .merge(address_type: type, mode: :lookup)
```
cintamani: I have changed the strategy here slightly so that the `uprn` is coming from the form as per standard-rails attribute and then is used to get the data as per before. The attributes are then sent to rails as normal strong parameters for mass-assignment

app/forms/waste_exemptions_engine/applicant_email_form.rb
```diff
@@ -13,9 +13,7 @@ def submit(params)
       # Assign the params for validation and pass them to the BaseForm method for updating
       self.confirmed_email = params[:confirmed_email]
 
-      attributes = { applicant_email: params[:applicant_email] }
-
-      super(attributes)
+      super(params.permit(:applicant_email))
```
cintamani: This will make mass-assignment to only operate on the `:applicant_email`, as the `permit` will override the previous `permit` which included `:confirmed_email`

app/forms/waste_exemptions_engine/contact_address_lookup_form.rb
```diff
@@ -10,10 +10,9 @@ class ContactAddressLookupForm < AddressLookupFormBase
     validates :contact_address, "waste_exemptions_engine/address": true
 
     def submit(params)
-      # Assign the params for validation and pass them to the BaseForm method for updating
-      contact_address = create_address(params[:contact_address][:uprn], :contact)
+      contact_address_attributes = get_address_data(params[:contact_address][:uprn], :contact)
```
cintamani: Since those forms are not actual `active_record` objects, the view is generating the names of this params using `contact_address` rather than `contact_address_attributes`. 
Since we have to populate those parameters anyway, and the code required for the form to behave like a model with the correct `accept_nested_attributes` defined methods is a long hack of delegation, I have decided to keep the params as they are now, and to rename the key to the correctly scoped `_attributes` when calling `super`. 
Note that the name of the key is important for the mass-assignment on the transient_registration to work.   
This trick is applied to other forms as well in this PR for the same reasons.

app/forms/waste_exemptions_engine/contact_address_manual_form.rb
```diff
@@ -2,62 +2,37 @@
...
-
-    def prefill_contact_address_params
-      return unless contact_address
 
-      self.premises = contact_address.premises&.strip
-      self.street_address = contact_address.street_address&.strip
-      self.locality = contact_address.locality&.strip
-      self.city = contact_address.city&.strip
-      self.postcode = contact_address.postcode&.strip
+      postcode == contact_address.postcode
```
cintamani: I finally figured that here instead of using `attr_accessor` for the address we can still do delegate as per the rest of the forms. The `attr_accessor` was still there as we have to keep an address stored BUT not replay it in case the user have gone back and changed the postcode. In order to achieve that same behaviour without the `attr_accessor` assignment of params, I have only kept the check for `address_still_valid` and set the address to `nil` in the in-memory transient_registration when that check is `false`. Same behaviour, but the code is in-line with the rest we have now.

app/forms/waste_exemptions_engine/exemptions_form.rb
```diff
@@ -7,10 +7,9 @@ class ExemptionsForm < BaseForm
     validates :exemptions, "waste_exemptions_engine/exemptions": true
 
     def submit(params)
-      # Rails authomatically delete params for which the value is empty :/
-      params ||= {}
+      params[:exemption_ids] ||= []
```
cintamani: This line here is in order to by-pass the Rails 4 bug for which empty parameters are parsed as `nil`. This is fixed in Rails 5, so whenever we upgrade we can also get rid of this

app/views/waste_exemptions_engine/exemptions_forms/new.html.erb
```diff
@@ -8,7 +8,7 @@
 
     <p><%= t(".description") %></p>
 
-    <div class="form-group <%= "form-group-error" if @exemptions_form.errors[:matched_exemptions].any? %>">
+    <div class="form-group <%= "form-group-error" if @exemptions_form.errors[:exemptions].any? %>">
```
cintamani: sneaking a minor visual bug fix in :D

app/forms/waste_exemptions_engine/exemptions_form.rb
Cruikshanks: I think it would help then to leave some comment to that affect. When this gets squashed the only place that will note that is this PR.

cintamani:
```suggestion
      # TODO: This line here is in order to by-pass the Rails 4 bug for which empty parameters are parsed as nil. 
      # This is fixed in Rails 5, so whenever we upgrade we can also get rid of this
      params[:exemption_ids] ||= []
```

app/forms/waste_exemptions_engine/exemptions_form.rb
```diff
@@ -7,10 +7,11 @@ class ExemptionsForm < BaseForm
     validates :exemptions, "waste_exemptions_engine/exemptions": true
 
     def submit(params)
-      # Rails authomatically delete params for which the value is empty :/
-      params ||= {}
+      # TODO: This line here is in order to by-pass the Rails 4 bug for which empty parameters are parsed as nil.
```
cintamani:
```suggestion
      # TODO: This line here is in order to by-pass the Rails 4 bug for which 
      # empty parameters are parsed as nil. 
```

app/forms/waste_exemptions_engine/exemptions_form.rb
```diff
@@ -7,10 +7,12 @@ class ExemptionsForm < BaseForm
     validates :exemptions, "waste_exemptions_engine/exemptions": true
 
     def submit(params)
-      # Rails authomatically delete params for which the value is empty :/
-      params ||= {}
+      # TODO: This line here is in order to by-pass the Rails 4 bug for which
```
cintamani:
```suggestion
      # TODO: This line here is in order to by-pass the Rails 4 bug for which
```

app/forms/waste_exemptions_engine/exemptions_form.rb
```diff
@@ -7,10 +7,12 @@ class ExemptionsForm < BaseForm
     validates :exemptions, "waste_exemptions_engine/exemptions": true
 
     def submit(params)
-      # Rails authomatically delete params for which the value is empty :/
-      params ||= {}
+      # TODO: This line here is in order to by-pass the Rails 4 bug for which 
+      # empty parameters are parsed as nil.
```
cintamani:
```suggestion
      # empty parameters are parsed as nil.
```

PR #390, opened 2019-10-07
Manual address forms delegate refactoring

cintamani:
Our manual addres forms have a couple of peculiarities, all to deal with the transient nature of the postcode.
Since the postcode is captured in a different screen, the address manual form has to decide wheter or not to show the in-store manual address depending on the fact that the postcode has changed or has remained the same. For this reson, in this cases, we are setting the attribute on the in-memory transient registration to `nil` wheter the postcode doesn't match the stored one, so that we can be consistent and delegate everything else.


cintamani:
Codeclimate is a won't fix

PR #399, opened 2019-10-30
Update address base url details

Cruikshanks:
In our local environment we hadn't quite got the configuration of the address base facade correct.

Turns out this was blocking vagrant from using the cassettes when running tests locally, because it could no longer match the requests.

This change fixes the issue.

### DIFF
--- a/spec/cassettes/postcode_no_matches.yml
+++ b/spec/cassettes/postcode_no_matches.yml
@@ -2,7 +2,7 @@
 http_interactions:
 - request:
     method: get
-    uri: http://localhost:9002/address-service/v1/addresses/postcode?client-id=0&key=client1&postcode=AA11AA
+    uri: http://localhost:3002/address-service/v1/addresses/postcode?client-id=0&key=client1&postcode=AA11AA
     body:
       encoding: US-ASCII
       string: ''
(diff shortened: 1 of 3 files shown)


spec/dummy/db/schema.rb
```diff
@@ -11,21 +11,11 @@
...
 
-  create_table "ad_renewal_letters_exports", force: :cascade do |t|
-    t.date     "expires_on"
-    t.string   "file_name"
-    t.integer  "number_of_letters"
-    t.string   "printed_by"
-    t.date     "printed_on"
-    t.datetime "created_at",        null: false
-    t.datetime "updated_at",        null: false
-  end
```
cintamani: I think we shouldn't kill this 🤔

Cruikshanks: I'm not, Rails is!

```bash
cd waste-exemptions-engine
git pull
bundle install
RAILS_ENV=test bundle exec rake db:drop
RAILS_ENV=test bundle exec rake db:create

cintamani: We should probably move that migration here?

Cruikshanks: No, I think it needs to stay where it is and this _does_ need to be deleted. Unless we have functionality in the engine that depends on this table existing (which based on our tests would suggest we don't as they pass with this missing from the DB) we shouldn't have it in the engine.

We've currently gone with a convention of keeping back office only related migrations in the back office (see https://github.com/DEFRA/waste-exemptions-back-office/tree/master/db/migrate). This sounds like it fits in that category.

PR #405, opened 2019-11-19
Use defra_ruby_address for address lookup

Cruikshanks:
One thing all our ruby digital services share is the need to lookup addresses based on a postcode. It's a need for users to enter an address as quickly and simply as possible. And it's a need for the organisation to capture as accurate an address as possible.

What this has led to though is a lot of duplication. So we have created the [Defra Ruby Address](https://github.com/DEFRA/defra-ruby-address) gem as a way of centralising and maintaining the code which performs the lookup in one place.

This change covers removing the existing lookup code, and switching to using the gem instead.


Cruikshanks:
Well, this has been a drawn-out side project!

Finally got the gem and everything finished off. Hopefully, this is another place we can start to reduce duplication and instead **reuse** code.

Plus we get a consistent way of handling responses with the `Response` object which we have used in other `defra_ruby` gems. And the cherry on the cake is not only does [defra-ruby-area](https://github.com/DEFRA/defra-ruby-area) support the lookups used by FRAE, WEX and WCR, it also supports the latest version of [ea-address-facade](https://github.com/DEFRA/ea-address-facade). So upgrading should now be a simple task.

lib/waste_exemptions_engine.rb
```diff
@@ -60,6 +61,13 @@ def use_last_email_cache
       change_string_to_boolean_for(@use_last_email_cache)
     end
 
+    def address_host=(value)
+      @address_host = value
```
irisfaraway: Why do we do this with the address host but not the Companies House host?

app/services/waste_exemptions_engine/determine_easting_and_northing_service.rb
```diff
@@ -34,27 +34,34 @@ def easting_and_northing_from_grid_reference(grid_reference)
...
-        default_do_not_fetch_again_coordinates
+      if response.successful?
+        @result[:easting] = response.results.first["x"].to_f
+        @result[:northing] = response.results.first["y"].to_f
 
         return
       end
 
-      no_result_from_postcode_lookup(postcode) && return if results.empty?
+      no_match_from_postcode_lookup(postcode) && return if response.error.is_a?(DefraRuby::Address::NoMatchError)
```
irisfaraway: I feel like this could all be cleaned up a bit to be easier to read, as at the moment we essentially have an `if/elsif/else` block but it's all structured differently. Maybe something like:

```
if response.successful?
  @result[:easting] = response.results.first["x"].to_f
  @result[:northing] = response.results.first["y"].to_f
elsif response.error.is_a?(DefraRuby::Address::NoMatchError)
  no_match_from_postcode_lookup(postcode)
else
  error_from_postcode_lookup(postcode, response.error)
end
```

Or maybe even some further refactoring, like:

```
if response.successful?
  set_result_coordinates_from_postcode_lookup(response)
elsif response.error.is_a?(DefraRuby::Address::NoMatchError)
  no_match_from_postcode_lookup(postcode)
else
  error_from_postcode_lookup(postcode, response.error)
end
```

What do you think?

Cruikshanks: I think I was trying to be too clever and use ruby idioms 😁 ! I think that makes a lot of sense.

lib/waste_exemptions_engine.rb
Cruikshanks: If we intend to keep the `attr_reader` then we really should be. I've chucked this together to try and illustrate

```ruby
class OtherConfig
  attr_accessor :host
end

class MainConfig
  attr_reader :host

  def initialize
    @other = OtherConfig.new
  end

  def host=(value)
    @other.host = value
  end
end

my_config = MainConfig.new
my_config.host = "example.com"
my_config.host
# nil

class FixedConfig
  attr_reader :host

  def initialize
    @other = OtherConfig.new
  end

  def host=(value)
    @host = value
    @other.host = value
  end
end

my_fixed_config = FixedConfig.new
my_fixed_config.host = "example.com"
my_fixed_config.host
# "example.com"
```

Essentially the `attr_reader :companies_house_host` is creating a hidden instance variable which we don't see and our custom setter isn't referencing when it comes to the companies house host.

So if you were to call it right now it would return `nil`. Putting the `@address_host = value` ensures a call to `foo.address_host` will return the value set.

But having figured this out with this PR, once merged we really should decide whether to bother with the attr_readers (so fix them) or delete them and just have the setters.

Ping me if this still doesn't make sense.

irisfaraway: Cool, makes sense to me :)

PR #427, opened 2020-02-03
Switch airbrake mgmt to defra-ruby-alert

Cruikshanks:
Whilst working on an issue for the [Flood Risk Activity Exemptions service](https://github.com/DEFRA/ruby-services-team/tree/master/services/frae) we came across an [issue](https://github.com/DEFRA/flood-risk-engine/pull/312) with Airbrake.

We wanted to add a new Rake task, and we wanted to apply our new pattern of ensuring Airbrake is closed when the task exits to ensure any errors are reported. The problem was due to the age of the Airbrake gem being used this caused an error in FRAE we hadn't seen in WEX and WCR.

So for reasons outlined in [FRAE PR 312](https://github.com/DEFRA/flood-risk-engine/pull/312) we came up with [defra-ruby-alert](https://github.com/DEFRA/defra-ruby-alert). The idea is going forward this will provide a single interface for configuring Airbrake and ensure the pattern of making sure all exceptions are reported before an app closes is applied consistently in all scenarios.

These changes are the first step in switching the WEX service to using it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,10 +11,3 @@ gemspec
 # your gemspec. These might include edge Rails or gems from your path or
 # Git. Remember to move these dependencies to your gemspec before releasing
 # your gem to rubygems.org.
-
-# TODO: Remove this and add a proper reference to defra_ruby_alert 1.0.0 in
-# the gemspec once we have confirmed it works with WEX and released it to
-# rubygems
-gem "defra_ruby_alert",
-    git: "https://github.com/DEFRA/defra-ruby-alert",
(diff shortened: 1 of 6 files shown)


Cruikshanks:
Also linked to https://github.com/DEFRA/waste-exemptions-back-office/pull/500 and https://github.com/DEFRA/waste-exemptions-front-office/pull/305 which are currently referring to this branch and have been updated to use defra_ruby_alert via the engine rather than config Airbrake directly.

PR #468, opened 2020-06-04
Update the project to use defra-ruby-alert latest

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-1060

The [defra-ruby-alert](https://github.com/DEFRA/defra-ruby-alert) gem needed to be updated to use the latest version of [Airbrake](https://github.com/airbrake/airbrake). It also needed some changes made to it because of that update to Airbrake.

Whilst we are still ironing out these issues we are holding off cutting a release of the gem. Hence this change updates the engine to use the Gemfile rather than the gemspec so we can source github directly.


Cruikshanks:
This work is already done 🤦 . Was looking at the wrong branch.

PR #483, opened 2020-07-07
Update validation gem

irisfaraway:
This was previously locked to the latest version (at the time). But we don't really need to lock it, and in fact would like to get the latest updates for Rails 6.


irisfaraway:
Once this is merged and rolled out to the host apps, https://eaflood.atlassian.net/browse/RUBY-1138 can be moved to QA.

PR #661, opened 2022-04-05
Playback operator address to use as site address

Beckyrose200:
https://eaflood.atlassian.net/browse/RUBY-1843

Here we are making changes so the user doesn't have to manually re-enter their site address if they have already entered it previously as either their business address or contact address. If the address is different from these ones then they have the opportunity to enter a new one or the grid reference. 

To Do:
~Write new tests for the new page request~
~Redo the routes testing to accommodate the new page~
I *think* these have been completed


spec/forms/waste_exemptions_engine/check_site_address_form_spec.rb
```diff
@@ -29,20 +29,46 @@ module WasteExemptionsEngine
...
+      end
+
+      context "when temp_reuse_address_for_site_location set to contact_address_option" do
+        let(:temp_reuse_address_for_site_location) { "contact_address_option" }
+
+        it "uses the contact address as the site address" do
+          subject
+
+          address_attributes.each do |attr|
+            expect(form.transient_registration.site_address.send(attr)).to eq(form.operator_address.send(attr))
```
PaulDoyle-DEFRA, line 60: Should compare site_address with contact_address, not with operator_address. Maybe vary the operator or contact address slightly to avoid having identical comparison targets.

PaulDoyle-DEFRA (changes requested):
Just one detail in the spec to be corrected.

spec/forms/waste_exemptions_engine/check_site_address_form_spec.rb
tobyprivett: Updated the factory to create different addresses and changed this spec to match the Contact address

PaulDoyle-DEFRA (approved):

PR #665, opened 2022-04-27
Companies House data retrieval 

Beckyrose200:
https://eaflood.atlassian.net/browse/RUBY-1872

Here we have the users company house number being used to validate that the company is active and return the company name and address for the user to confirm it is indeed their company. If the user confirms this then the name name registered on company's house will be used as the operator name for the service. If the user doesn't confirm these details then they are asked to change them via company house


app/controllers/waste_exemptions_engine/incorrect_companies_house_details_forms_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module WasteExemptionsEngine
+  class IncorrectCompaniesHouseDetailsFormsController < FormsController
```
Beckyrose200, line 4: @PaulDoyle-DEFRA Is the naming of incorrect_companies_house_details okay? Wasn't sure if it was correct or not, can change the name if need to :)

PaulDoyle-DEFRA: It's a bit of a mouthful. Maybe use `incorrect_company_forms_controller` like in WCR?

Beckyrose200: Yeah that sounds better, will make the changes now, thanks! :)

app/forms/waste_exemptions_engine/check_registered_name_and_address_form.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module WasteExemptionsEngine
+  class CheckRegisteredNameAndAddressForm < BaseForm
+    delegate :temp_use_registered_company_details, to: :transient_registration
+    delegate :company_no, to: :transient_registration
+    delegate :registered_company_name, to: :transient_registration
+    delegate :operator_name, to: :transient_registration
```
PaulDoyle-DEFRA, line 8: There is no operator_name attribute on transient_registration. Presumably this isn't used anyway.

spec/models/waste_exemptions_engine/new_registration_workflow_state/operator_name_form_spec.rb
```diff
@@ -14,29 +14,6 @@ module WasteExemptionsEngine
           expect(new_registration).to transition_from(current_state).to(next_state).on_event(:next)
         end
 
-        context "when neither new_registration.partnership? nor " \
```
PaulDoyle-DEFRA, line 17: Has this test logic moved somewhere else?

PaulDoyle-DEFRA:
Gemfile.lock has changed since you created your branch, please rebase to main.

PaulDoyle-DEFRA (changes requested):
Please rebase to main

Beckyrose200:
@PaulDoyle-DEFRA I had done a rebase earlier today, tried to do one again now but it says its all up to date. I forced pushed it and thats not showing up either because there is no change. Whats the difference between the Gemlock files?

app/forms/waste_exemptions_engine/check_registered_name_and_address_form.rb
PaulDoyle-DEFRA: As discussed - this is fine as it is, my understanding of this pattern needs work.

spec/models/waste_exemptions_engine/new_registration_workflow_state/operator_name_form_spec.rb
PaulDoyle-DEFRA: As discussed - this is fine, the specs that were removed are no longer valid and the ones that remain are all that is now needed.

PaulDoyle-DEFRA:
As discussed - ignore my comment about rebasing, wrong repo...

PaulDoyle-DEFRA (dismissed):

PR #671, opened 2022-05-12
[1875] Back Office change 'operator name' edit button state

d-a-v-e:
https://eaflood.atlassian.net/browse/RUBY-1875


app/views/waste_exemptions_engine/edit_forms/new.html.erb
```diff
@@ -100,15 +100,19 @@
...
+            <%= name_label(@edit_form) %>
           </dt>
           <dd class="govuk-summary-list__value">
             <%= @edit_form.operator_name %>
           </dd>
           <dd class="govuk-summary-list__actions">
-            <%= link_to operator_name_edit_forms_path(@edit_form.token) do %>
-              <%= t(".edit_links.default") %>
-              <span class="govuk-visually-hidden"> <%= t(".edit_links.visually_hidden.operator_name") %></span>
+            <% if %w[limitedCompany limitedLiabilityPartnership].include?(@edit_form.business_type) %>
```
PaulDoyle-DEFRA, line 109: There are existing helpers for this type of logic, e.g. `company_no_required?` in `can_have_registration_attributes`.

config/locales/forms/edit_forms/en.yml
```diff
@@ -29,7 +29,7 @@ en:
           operator:
             heading: "Operator details"
             labels:
-              operator_name: "Name"
+              operator_name: "Registered Name"
```
PaulDoyle-DEFRA, line 32: @Beckyrose200 is making similar changes to another locale file, please liaise for consistency.

PaulDoyle-DEFRA (changes requested):
Please reinstate 18n and add unit tests to resolve the SonarCloud complaint

app/helpers/waste_exemptions_engine/edit_helper.rb
```diff
@@ -2,6 +2,8 @@
 
 module WasteExemptionsEngine
   module EditHelper
+    LOCALE = "waste_exemptions_engine.edit_forms"
```
PaulDoyle-DEFRA, line 5: Why does the i18n scope need to be added explicitly?

app/models/concerns/waste_exemptions_engine/can_have_registration_attributes.rb
```diff
@@ -15,6 +15,10 @@ module CanHaveRegistrationAttributes
 
     # Some business types should not have a company_no
     def company_no_required?
+      llp_or_ltd?
```
PaulDoyle-DEFRA, line 18: Is this change so you can use `llp_or_ltd?` as an alias for `company_no_required?`

app/helpers/waste_exemptions_engine/edit_helper.rb
d-a-v-e: It's referenced a couple times in later on so just defining it here. What would you suggest instead?

app/models/concerns/waste_exemptions_engine/can_have_registration_attributes.rb
d-a-v-e: yep - haven't aliased `company_no_required` incase that were to change

app/helpers/waste_exemptions_engine/edit_helper.rb
PaulDoyle-DEFRA: We don't need to state the scope explicitly elsewhere, wondering why that's needed here. Is it because it's a helper?

app/models/concerns/waste_exemptions_engine/can_have_registration_attributes.rb
PaulDoyle-DEFRA: `company_no_required?` is used elsewhere in the codebase to make this determination. It's not clear that adding another term for the same thing helps. I think the key information here is not whether the business is ltd or llp, but whether the user journey needs to deal with companies house number, so `company_no_required?` is probably a reasonable name to use. But clearly you didn't find it intuitive, so it's not as intuitive as I thought - maybe need to discuss this :-)

app/helpers/waste_exemptions_engine/edit_helper.rb
d-a-v-e: if you look further down I make use of the LOCALE in the call to i18n - given the scope of the helper is only for edit forms I've defined the scope at the top so it can be used wherever the i18n is called. If it looks odd I can just repeatedly write `waste_exemptions_engine.edit_forms` in the `t()` calls instead?

app/models/concerns/waste_exemptions_engine/can_have_registration_attributes.rb
d-a-v-e: I've gone only by the ticket's requirement which is "only make this change if it's a LTD or LLP" - if that's synonymous with needing a company number then happy to rollback just give the word 👍

PaulDoyle-DEFRA: Yes, that's correct, "company number is required" and "business type is LTD or LLP" are the same thing.

app/helpers/waste_exemptions_engine/edit_helper.rb
d-a-v-e: @PaulDoyle-DEFRA this is resolved now - I can leave off the prefix but need to add it in the tests like so: https://github.com/DEFRA/waste-exemptions-engine/pull/671/files#diff-b8714c1d7299d1d3d6b05c2bda31f96e36cc9b25e996aadbbbc1a978fae7a34aR7

PaulDoyle-DEFRA (approved):

PR #672, opened 2022-05-13
New registration relabel operator name on check answers page

Beckyrose200:
https://eaflood.atlassian.net/browse/RUBY-1873
This is a label change for the check your answers page. When the business type is either a LLP or company the label is changed to Registered Name (as this is pulled from companies house and isn't entered by the user) however if its anything else i.e. charity then the label is Operator Name. 

Spec file has been changed to test the new method and to test a charity on the 'check your answers page' as well.


app/models/concerns/waste_exemptions_engine/can_use_new_registration_workflow.rb
```diff
@@ -322,6 +322,10 @@ module CanUseNewRegistrationWorkflow
                       to: :check_registered_name_and_address_form,
                       unless: :skip_registration_number?
 
+          transitions from: :operator_postcode_form,
```
PaulDoyle-DEFRA, line 325: This is unrelated to the change as described in the PR description. Was it included accidentally?

Beckyrose200: Its been duplicated by the looks of it, I rebased the branch which had quite a lot of changes and I think its duplicated itself when I was accepting the current changes. Will remove this now :)

PaulDoyle-DEFRA (approved):

PR #679, opened 2022-06-08
Standard error raised for failing company name

Beckyrose200:
https://eaflood.atlassian.net/browse/RUBY-1877 
Here we have added some logic in the refresh company house service to throw an error if the company name is returning nil.


app/services/waste_exemptions_engine/refresh_companies_house_name_service.rb
```diff
@@ -7,6 +7,8 @@ class RefreshCompaniesHouseNameService < WasteExemptionsEngine::BaseService
     def run(company_reference)
       registration = Registration.find_by(reference: company_reference)
       company_name = DefraRubyCompaniesHouse.new(registration.company_no).company_name
+      raise StandardError if company_name == nil?
```
PaulDoyle-DEFRA, line 10: Use `.new` and give the exception a message, e.g: `raise StandardError.new "Company name not found"`, or just `raise "Company name not found"`

app/services/waste_exemptions_engine/refresh_companies_house_name_service.rb
```diff
@@ -7,6 +7,8 @@ class RefreshCompaniesHouseNameService < WasteExemptionsEngine::BaseService
     def run(company_reference)
       registration = Registration.find_by(reference: company_reference)
       company_name = DefraRubyCompaniesHouse.new(registration.company_no).company_name
```
PaulDoyle-DEFRA, line 9: What type of exception does RestClient raise if the URL is invalid or the request times out? Need to handle that as well.

PaulDoyle-DEFRA (changes requested):
Need to handle other exception types including timeouts and URL errors, possibly in DefraRubyCompaniesHouse.

lib/defra_ruby_companies_house.rb
```diff
@@ -7,7 +7,7 @@ def initialize(company_no)
     @company_url = "#{Rails.configuration.companies_house_host}#{company_no.to_s.rjust(8, '0')}"
     @api_key = Rails.configuration.companies_house_api_key
 
-    load_company
+    StandardError unless load_company
```
PaulDoyle-DEFRA, line 10: `raise StandardError unless load_company`

spec/lib/defra_ruby_companies_house_spec.rb
```diff
@@ -70,4 +70,14 @@
       end
     end
   end
+
+  context "when the service is down" do
+    before do
+      allow_any_instance_of(DefraRubyCompaniesHouse).to receive(:load_company).and_raise(StandardError)
```
PaulDoyle-DEFRA, line 76: I think this should be raising one of the RestClient exception types - the test is to confirm that it get mapped to StandardError

lib/defra_ruby_companies_house.rb
Beckyrose200: image
Putting raise in makes a lot of the rspec tests fail

spec/lib/defra_ruby_companies_house_spec.rb
PaulDoyle-DEFRA: Actually there should be one test each for (a) when the company number is not valid (ResourceNotFound) and (b) when the service is down (NotFound)

PaulDoyle-DEFRA (changes requested):
Apologies for noticing earlier, but I think RefreshCompaniesHouseNameService should be in the back-office, not the engine, because it is back-office only functionality.

PaulDoyle-DEFRA (approved):

PR #687, opened 2022-07-14
Send letters to contacts with no email address

Beckyrose200:
https://eaflood.atlassian.net/browse/RUBY-1997

Here we have the changes to allow letters to be sent to registrations without an email address. 
In WEX there are two emails address captured, applicant_email and contact_email. 
These changes mean letters are sent to any emails that are nil value instead of the NCCC email address.

A plural helper method was also affected with the change, and has been updated to allow for a nil address.

### DIFF
--- a/app/helpers/waste_exemptions_engine/plurals_helper.rb
+++ b/app/helpers/waste_exemptions_engine/plurals_helper.rb
@@ -6,6 +6,7 @@ module WasteExemptionsEngine
       emails = [form.applicant_email, form.contact_email]
       emails.uniq!
       emails.delete(nil)
+      emails.delete("")
 
       if emails.empty?
         "letter"
--- a/app/helpers/waste_exemptions_engine/plurals_helper.rb
+++ b/app/helpers/waste_exemptions_engine/plurals_helper.rb
@@ -5,7 +5,7 @@ module WasteExemptionsEngine
     def confirmation_comms(form)
       emails = [form.applicant_email, form.contact_email]
       emails.uniq!
-      emails.delete(WasteExemptionsEngine.configuration.assisted_digital_email)
+      emails.delete(nil)
 
       if emails.empty?
         "letter"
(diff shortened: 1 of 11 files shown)


Beckyrose200:
Need to look at options for notify template if applicant email isn't present

app/helpers/waste_exemptions_engine/plurals_helper.rb
```diff
@@ -5,7 +5,7 @@ module PluralsHelper
     def confirmation_comms(form)
       emails = [form.applicant_email, form.contact_email]
       emails.uniq!
-      emails.delete(WasteExemptionsEngine.configuration.assisted_digital_email)
+      emails.delete(nil)
```
PaulDoyle-DEFRA, line 8: Does this also need to handle blank emails, i.e. empty strins values?

config/locales/confirmation_letters.en.yml
```diff
@@ -29,6 +29,7 @@ en:
       position: "Position:"
       telephone: "Telephone:"
       email: "Email:"
+      no_email: "Not Present"
```
PaulDoyle-DEFRA, line 32: Possibly needs to be lower-case `P` for consistency

app/helpers/waste_exemptions_engine/plurals_helper.rb
Beckyrose200: I don't think so, because you can't enter a blank string or it will give a validation error when your entering an email. So with that do I still need to include it as a possibility?

PaulDoyle-DEFRA: `contact_email` is being set to an empty string in WCR using similar code. However the WCR code handles both blank and nil email addresses, so it's more something to tidy up rather than a defect - Tim has created https://eaflood.atlassian.net/browse/RUBY-2028 to fix this later. I don't think we can rule out blank emails here, so best allow for the possibility in the code which refers to the email address attributes.

PaulDoyle-DEFRA (approved):

PR #688, opened 2022-07-15
Refactor back-navigation

PaulDoyle-DEFRA:
This change mirrors recent changes to the waste-carriers-engine to simplify the back-navigation mechanism without modifying the user experience. It does this by maintaining an ordered list of previous forms visited, adding to the list as the user progresses through the forms. When the user navigates back, they are brought to the last form in the list, which is then removed from the list.

In the event that there is an invalid entry in the list, that entry will be skipped, so the user could in theory be brought back more than one step in the flow. If no valid entry is found in the list, they will be brought back to the start form. This is assumed to be acceptable as it should only happen in rare cases, e.g. changes to form navigation impacting a transient registration that was created before the deployment of the change and continued afterwards.

Note that the bulk of the changes are for the removal of the existing hard-wired back-navigation unit tests. The key files for the new mechanism are:
* `transient_registration.rb`
* `forms_controller.rb`
* `_back.html.erb`
* `routes.rb`
https://eaflood.atlassian.net/browse/RUBY-1986


spec/dummy/db/schema.rb
```diff
@@ -10,11 +10,22 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_07_08_103727) do
+ActiveRecord::Schema.define(version: 2022_07_14_164313) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "ad_renewal_letters_exports", id: :serial, force: :cascade do |t|
```
Beckyrose200, line 18: Think Ive had previous issues with ad_renew_letters_exports, is this meant to be in the schema? :)

Beckyrose200 (approved):
Looks really good 😊

spec/dummy/db/schema.rb
PaulDoyle-DEFRA: Good catch, I must have re-added it while trying to resolve a local migration issue. Removed now.

PR #721, opened 2022-10-28
Set renewal expiry date

PaulDoyle-DEFRA:
This change sets the expiry date for renewals to three years from original registration expiry date, as opposed to the existing value of three years from the date of the renewal.
https://eaflood.atlassian.net/browse/RUBY-2167


PaulDoyle-DEFRA:
@timstone - I've confirmed that the lines about which SonarCloud is complaining are executed during unit test, not clear why it's flagging them for lack of coverage. On that basis I'll push the PR through once approved.

app/models/concerns/waste_exemptions_engine/can_activate_exemption.rb
```diff
@@ -22,7 +22,15 @@ module CanActivateExemption
       # Transition effects
       def activate_exemption
         self.registered_on = Date.today
-        self.expires_on = Date.today + (WasteExemptionsEngine.configuration.years_before_expiry.years - 1.day)
+
+        self.expires_on =
+          if transient_registration.is_a? WasteExemptionsEngine::RenewingRegistration
+            transient_registration.registration.expires_on +
+              WasteExemptionsEngine.configuration.years_before_expiry.years
```
timstone, line 29: Hard to get my head round this but should this be 3 years minus one day like in line 31? so it expires on the last day

PaulDoyle-DEFRA: a/c says: "the new registration will have an expiry date, three years after the current registrations expiry date"

timstone (approved):

timstone (dismissed):

timstone (approved):

PR #722, opened 2022-10-28
site postcode error page

PaulDoyle-DEFRA:
This change removes the manual site address option on the site postcode error page.
https://eaflood.atlassian.net/browse/RUBY-2120


timstone (commented):
Shouldn't it have a link back to `Use a grid reference instead` ?

config/locales/forms/site_postcode_forms/en.yml
```diff
@@ -8,16 +8,17 @@ en:
...
       models:
         waste_exemptions_engine/site_postcode_form:
           attributes:
             temp_site_postcode:
-              blank: "Enter a postcode"
-              wrong_format: "Enter a valid UK postcode"
-              no_results: "We cannot find any addresses for that postcode. Check the postcode or enter the address manually."
+              blank: Enter a postcode
+              wrong_format: Enter a valid UK postcode
+              no_results: We cannot find any addresses for that postcode. Check the postcode or enter the address manually.
```
timstone, line 21: Appreciate there is no wireframe for this scenario and Will is off. 
`Check the postcode or enter the address manually.` will need to change to something like
`Check the postcode or use a grid reference instead.`

timstone (changes requested):
One content change

timstone (approved):

PR #877, opened 2023-10-18
Fix/ruby 2728 exemption status on cert

PaulDoyle-DEFRA:
Show active and de-registered exemptions with status on PDF certificate.
https://eaflood.atlassian.net/browse/RUBY-2728

### DIFF
--- a/app/models/concerns/waste_exemptions_engine/can_iterate_exemptions.rb
+++ b/app/models/concerns/waste_exemptions_engine/can_iterate_exemptions.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module WasteExemptionsEngine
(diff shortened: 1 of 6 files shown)


app/models/concerns/waste_exemptions_engine/can_iterate_exemptions.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module WasteExemptionsEngine
+  module CanIterateExemptions
```
jjromeo, line 4: I'm not quite understanding the name of this concern, would something like CanSortExemptions be a bit clearer?

jjromeo (approved):

PR #1011, opened 2024-06-11
Feature/ruby 3121 cya on farm

PaulDoyle-DEFRA:
This adds a change flor for the "on a farm" element of the check answers page.
https://eaflood.atlassian.net/browse/RUBY-3121


config/locales/partials/data_overview/registration_info.en.yml
```diff
@@ -12,6 +12,7 @@ en:
             value:
               "true": "Yes"
               "false": "No"
+            change_link_suffix: "on a farm"
```
brujeo, line 15: According to AC#3, this should be "Will this operation be carried out on a farm"

brujeo (approved):

PR #1485, opened 2026-04-20
[RUBY-4229] Add England-only site location restriction feature

jjromeo:
https://eaflood.atlassian.net/browse/RUBY-4229

Adds a front-office, feature-toggled England-only restriction for site location flows in the app.

  When `restrict_site_locations_to_england` is enabled:

  - Grid references outside England are rejected
  - Postcodes with no English matches are rejected
  - Address lookup results are filtered to English addresses only
  - If EA area lookup errors, the user is allowed to continue

  ## Changes

  - Adds CheckSiteLocationIsInEnglandService to do the England check
  - The restriction / validations are added in to:
      - SiteGridReferenceForm
      - SitePostcodeForm
      - SiteAddressLookupForm
  - scoped the behaviour to front office only via host_is_back_office?
  - added README documentation for the local EA area lookup flow and where coordinates come from

  ## Notes

  - EA area lookup uses the local ea_public_face_areas table
  - postcode address lookup remains external


app/forms/concerns/waste_exemptions_engine/can_restrict_site_locations_to_england.rb
```diff
@@ -0,0 +1,22 @@
...
+  module CanRestrictSiteLocationsToEngland
+    private
+
+    def restrict_site_locations_to_england?
+      FeatureToggle.active?(:restrict_site_locations_to_england) &&
+        !WasteExemptionsEngine.configuration.host_is_back_office?
+    end
+
+    def site_location_in_england?(grid_reference: nil, easting: nil, northing: nil)
+      return true unless restrict_site_locations_to_england?
```
brujeo, line 13: This line feels contradictory - when restrictions are disabled, we’re asserting that the location is in England without checking.

app/models/waste_exemptions_engine/ea_public_face_area.rb
```diff
@@ -7,6 +7,7 @@ class EaPublicFaceArea < ApplicationRecord
     validates :code, presence: true
 
     OUTSIDE_ENGLAND_CODE = "OUTSIDE_ENGLAND"
+    OUTSIDE_ENGLAND_NAME = "Outside England"
```
brujeo, line 10: Shouldn't the name be in translation files?

app/forms/waste_exemptions_engine/site_grid_reference_form.rb
```diff
@@ -86,5 +88,15 @@ def update_existing_site
 
       true
     end
+
+    def check_grid_reference_location?
+      grid_reference.present?
```
brujeo, line 93: should this also be checking for restrict_site_locations_to_england? ?

app/services/waste_exemptions_engine/check_site_location_is_in_england_service.rb
```diff
@@ -0,0 +1,34 @@
...
+      return true unless valid_coordinates?(coordinates[:easting], coordinates[:northing])
+
+      area = DetermineAreaService.run(
+        easting: coordinates[:easting],
+        northing: coordinates[:northing]
+      )
+
+      area != EaPublicFaceArea::OUTSIDE_ENGLAND_NAME
+    rescue StandardError
+      true
```
brujeo: Shouldn't we log the error?

app/forms/waste_exemptions_engine/site_grid_reference_form.rb
jjromeo: No this is just deciding whether to validate it

app/models/waste_exemptions_engine/ea_public_face_area.rb
jjromeo: No because it shouldn't be translated

brujeo (approved):

PR #1501, opened 2026-05-20
implement message delivery status call back

brujeo:
WEX - Implement message delivery status call back
https://eaflood.atlassian.net/browse/RUBY-4284

- Add index to communication_logs on notification_id
- Implement message delivery status callback and update communication log
- Add NotifyCallbackJob and corresponding specs for message delivery status handling
- Implement NotifyCallbacksController and associated routes for handling GOV.UK Notify callbacks

### DIFF
--- a/app/controllers/waste_exemptions_engine/notify_callbacks_controller.rb
+++ b/app/controllers/waste_exemptions_engine/notify_callbacks_controller.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+module WasteExemptionsEngine
+  class NotifyCallbacksController < ::WasteExemptionsEngine::ApplicationController
+    protect_from_forgery with: :null_session
+
+    def process_callback
+      validate_bearer_token!
+
+      request.body.rewind
+      body = request.body.read
+      payload = JSON.parse(body)
+
+      NotifyCallbackJob.perform_later(payload)
+    rescue StandardError => e
+      Rails.logger.error "Notify callback error: #{e}"
+      Airbrake.notify(e)
+    ensure
+      head :ok
+    end
+
+    private
+
+    def validate_bearer_token!
+      expected_token = ENV.fetch("NOTIFY_CALLBACK_BEARER_TOKEN", nil)
+      raise "Notify callback bearer token not configured" if expected_token.blank?
+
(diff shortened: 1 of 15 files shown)


app/controllers/waste_exemptions_engine/notify_callbacks_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    class TokenNotConfiguredError < StandardError; end
+    class MissingAuthorizationError < StandardError; end
+    class InvalidTokenError < StandardError; end
+
+    protect_from_forgery with: :null_session
+
+    def process_callback
+      validate_bearer_token!
+
+      request.body.rewind
```
jjromeo, line 15: Why is it necessary to rewind the body before it has been read?

app/services/waste_exemptions_engine/notify_callback_handler_service.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+module WasteExemptionsEngine
+  class NotifyCallbackHandlerService < BaseService
+    NOTIFICATION_STATUSES = %w[delivered permanent-failure technical-failure returned].freeze
```
jjromeo: Should the fact that they are terminal ones should that be included in the constant name?

app/controllers/waste_exemptions_engine/notify_callbacks_controller.rb
brujeo: It's a common pattern to prevent from empty-string bug in case request body has been already read. 
https://github.com/DEFRA/waste-carriers-engine/blob/e882b3b4f044a1d9cc342f981ff59904704a9163/app/controllers/waste_carriers_engine/govpay_webhook_callbacks_controller.rb#L12

brujeo: I've added a comment for clarity

app/services/waste_exemptions_engine/notify_callback_handler_service.rb
brujeo: renamed

jjromeo (approved):

PR #1502, opened 2026-05-21
[RUBY-4260] Add multisite support for registration edit link email service

jjromeo:
https://eaflood.atlassian.net/browse/RUBY-4260


app/services/waste_exemptions_engine/registration_edit_link_email_service.rb
```diff
@@ -33,23 +33,36 @@ def communications_log_params
     private
 
     def template_id
```
brujeo, line 35: So, template_id is now dynamic, but the **template_label** in communication_log_params remains static and will log it as "Registration edit link email" regardless of whether it was multi-site or not

brujeo (approved):

PR #1513, opened 2026-06-05
[RUBY-4312] Refactor site address handling to use SaveSiteAddressService

jjromeo:
https://eaflood.atlassian.net/browse/RUBY-4312

Fix bug with duplicate sites


app/services/waste_exemptions_engine/save_site_address_service.rb
brujeo, line 1: I don’t see a detailed spec for SaveSiteAddressService.
It’s important that we unit-test it in isolation and cover all possible edge cases.

brujeo (approved):

